PR #2, opened 2020-08-19
Change persistence backend to RDS

danwilliams:
## Summary of changes

The persistence backend has been changed to Amazon Aurora's MySQL RDS backend. The backend is communicated with using boto3's  `rds-data` client which executes SQL statements over HTTPS.

The backend database itself is produced by a separate repository, (`covid-engineering`). That repository's pipeline produces several dev and production database instances, that this app can integrate with.

Additional configuration variables are required; existing and new users are advised to view the readme in order to set up the changes.

There are the representation of enums switched from the legacy representation which used strings, to an int representation. A substantial amount of the code change is focused on that switch.

Note: As of the latest commit, this branch has a dependency on https://github.com/alphagov/covid-engineering/pull/682. Until that is merged, retrieving results from RDS won't work.

## Testing instructions

An AWS account is needed, as well as access to the Amazon Aurora MySQL database produced by the `covid-engineering` pipeline. By following the README and providing suitably privileged AWS credentials it should then be possible to go through the full form journey.


vulnerable_people_form/integrations/persistence.py
```diff
@@ -0,0 +1,203 @@
...
+
+def get_secretsmanager_client(app=current_app):
+    return boto3.client(
+        "secretsmanager", endpoint_url=app.config.get("LOCAL_AWS_ENDPOINT_URL")
+    )
+
+
+def _find_database_arn(app):
+    rds = get_rds_client(app=app)
+    clusters = rds.describe_db_clusters()
```
sanisc, line 24: this should be    ` clusters = rds.describe_db_clusters()['DBClusters']` as its the clusters information needed

vulnerable_people_form/integrations/persistence.py
```diff
@@ -0,0 +1,203 @@
...
+        app.config["AWS_RDS_DATABASE_ARN"] = _find_database_arn(app)
+
+    if app.config.get("AWS_RDS_SECRET_ARN_OVERRIDE"):
+        app.config["AWS_RDS_SECRET_ARN"] = app.config["AWS_RDS_SECRET_ARN_OVERRIDE"]
+    else:
+        app.config["AWS_RDS_SECRET_ARN"] = _find_database_secret_arn(app)
+
+
+def _form_response_database_name():
+    return current_app.config.get("AWS_RDS_DATABASE", "coronavirus-vulnerable-people")
```
sanisc, line 53: general question - where is this value from  for the RDS DB?

README.md
```diff
@@ -202,9 +217,23 @@ variables.
     have the form
     `http(s)://<externally_reachable_app_url>/nhs-login-callback`.
 
-  - `AWS_DYNAMODB_SUBMISSIONS_TABLE_NAME` **\[required\]**: This
-    variable contains the table name that the app will submit form
-    responses to.
+  - `AWS_RDS_DATABASE_NAME` **\[required\]**: This variable sets the
+    database that the app will use to persist form answers.
+
+  - `AWS_RDS_DATABASE_SCHEMA` **\[required\]**: this variable sets the
```
sanisc, line 223: can this be blank for prod?

vulnerable_people_form/integrations/persistence.py
danwilliams: Thanks! 🙂

vulnerable_people_form/integrations/persistence.py
danwilliams: This method is no longer used. Removed.

README.md
danwilliams: It can be set to a blank string, but it is currently set as a required variable, which is not super-logical! Changed.

vulnerable_people_form/integrations/persistence.py
```diff
@@ -0,0 +1,199 @@
...
+def generate_date_parameter(name, value):
+    return {
+        "name": name,
+        "typeHint": "DATE",
+        "value": {"stringValue": "{year}-{month:0>2}-{day:0>2}".format(**value),},
+    }
+
+
+def execute_sql(sql, parameters):
+    return get_rds_data_client().execute_statement(
```
sanisc, line 82: The SQL mode needs to be set to strict. e.g
`SET SESSION sql_mode = 'STRICT_ALL_TABLES';`

vulnerable_people_form/form_pages/shared/answers_enums.py
```diff
@@ -1,44 +1,76 @@
...
+    }
 
 
 def get_radio_options_from_enum(target_enum, selected_value):
     return [
         {
             "value": value.value,
-            "text": value.value,
+            "text": value.value_as_text,
             "checked": selected_value == value.value,
```
sanisc, line 74: trailing comma should be removed

vulnerable_people_form/form_pages/shared/answers_enums.py
```diff
@@ -1,44 +1,76 @@
...
-class MedicalConditionsAnswers(enum.Enum):
-    YES = "Yes, I have one of the listed medical conditions"
-    NO = "No, I do not have one of the listed medical conditions"
+class MedicalConditionsAnswers(EnumWithText):
+    YES = 1
+    NO = 0
+
+    __text_values__ = {
+        YES: "Yes, I have one of the listed medical conditions",
+        NO: "No, I do not have one of the listed medical conditions",
```
sanisc, line 65: trailing comma should be removed

vulnerable_people_form/form_pages/shared/answers_enums.py
```diff
@@ -1,44 +1,76 @@
...
-    NOT_SURE = "Not sure"
+class NHSLetterAnswers(EnumWithText):
+    YES = 1
+    NO = 2
+    NOT_SURE = 3
+
+    __text_values__ = {
+        YES: "Yes, I’ve been told to shield",
+        NO: "No, I haven’t been told to shield",
+        NOT_SURE: "Not sure",
```
sanisc, line 54: trailing comma should be removed

vulnerable_people_form/form_pages/shared/answers_enums.py
```diff
@@ -1,44 +1,76 @@
...
-class ApplyingOnOwnBehalfAnswers(enum.Enum):
-    YES = "Yes, I'm applying on my own behalf."
-    NO = "No, I'm applying on behalf of someone else."
+class ApplyingOnOwnBehalfAnswers(EnumWithText):
+    YES = 1
+    NO = 0
+
+    __text_values__ = {
+        YES: "Yes, I'm applying on my own behalf.",
+        NO: "No, I'm applying on behalf of someone else.",
```
sanisc, line 30: trailing comma should be removed

vulnerable_people_form/form_pages/dietary_requirements.py
```diff
@@ -8,7 +8,7 @@
     form_answers,
     get_errors_from_session,
     request_form,
-    update_session_answers_from_form,
+    update_session_answers_from_form_for_enum,
```
sanisc, line 11: trailing comma should be removed - probably should go through all the files in the commit to double check this as i have seen it in quiet a few places

README.md
sanisc: Might be best to remove this as its not supported 
https://docs.aws.amazon.com/rdsdataservice/latest/APIReference/API_ExecuteStatement.html

vulnerable_people_form/integrations/persistence.py
```diff
@@ -0,0 +1,199 @@
...
+        sql=sql,
+        parameters=parameters,
+        database=current_app.config["AWS_RDS_DATABASE_NAME"],
```
sanisc, line 85: i don't think the DB name is required and its a value that i don't think has been provided. The connection can work of the secret and the resourcearn

vulnerable_people_form/integrations/postcode_eligibility.py
```diff
@@ -1,6 +1,15 @@
 import os
 
+from .persistence import get_rds_data_client, generate_string_parameter, execute_sql
+
 
 def check_postcode(postcode):
-    # TODO: add implementation
-    return os.environ.get("FAIL_POSTCODE_VERIFICATION") is None
+    records = execute_sql(
+        "CALL is_postcode_on_lockdown(:postcode)",
```
sanisc, line 8: i think this should be  "CALL cv_base.is_postcode_on_lockdown(:postcode)"

vulnerable_people_form/form_pages/dietary_requirements.py
danwilliams: The trailing comma is part of general good practice standards.

sanisc: I have tended to see it used in collections or lists but no so much else where either thats fine

README.md
danwilliams: Good spot 👍

vulnerable_people_form/integrations/persistence.py
danwilliams: The Aurora HTTP endpoint only accepts single statements. What we could do is open a transaction, get a transaction id, execute the two statements in two separate HTTP requests, then issue a 4th HTTP request to close the transaction. However, this seems rather a lot of effort.

Would it be better to set `sql_mode` at the instance_level instead, rather than relying on clients to implement correct behaviour?

https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/AuroraMySQL.Reference.html#AuroraMySQL.Reference.ParameterGroups

vulnerable_people_form/integrations/persistence.py
danwilliams: Good point! Updated 👍

vulnerable_people_form/integrations/postcode_eligibility.py
danwilliams: It appears that it has been silently renamed in the relevant `covid-engineering` PR, although not in the PR description. As the PR is not yet merged, this was not known before, as we only had the "it will be like this" info given previously. Updated! 🙂

vulnerable_people_form/integrations/persistence.py
```diff
@@ -0,0 +1,199 @@
...
+    if app.config.get("AWS_RDS_DATABASE_ARN_OVERRIDE"):
+        app.config["AWS_RDS_DATABASE_ARN"] = app.config["AWS_RDS_DATABASE_ARN_OVERRIDE"]
+    else:
+        app.config["AWS_RDS_DATABASE_ARN"] = _find_database_arn(app)
+
+    if app.config.get("AWS_RDS_SECRET_ARN_OVERRIDE"):
+        app.config["AWS_RDS_SECRET_ARN"] = app.config["AWS_RDS_SECRET_ARN_OVERRIDE"]
+    else:
+        app.config["AWS_RDS_SECRET_ARN"] = _find_database_secret_arn(app)
+
```
thomasridd, line 50: The purpose of all the `_find_..._arn()` complications is that the arn values are capable of changing without warning

Have you got a fallback anywhere to refresh these values if the job fails?

vulnerable_people_form/integrations/persistence.py
samcrang: Yeah, we're working on setting `sql_mode` at the global level. There's a branch with it being worked on over here https://github.com/alphagov/covid-engineering/compare/enable-mysql-strict-mode#diff-a36f48a6afdedf2f324409d542bf5c82

I think at the moment I'd rather just do the four requests until we can get the setting of the global `sql_mode` done, then we can undo this work if it turns out it's too slow.

My preference is that I'd rather it be slower but less likely to allow bad data to be ingested (until we get the global option set).

vulnerable_people_form/integrations/persistence.py
```diff
@@ -0,0 +1,199 @@
...
+    have_you_received_an_nhs_letter,
+    do_you_want_supermarket_deliveries,
+    do_you_need_help_meeting_your_basic_care_needs,
+    do_you_have_any_special_dietary_requirements,
+    do_you_have_someone_in_the_house_to_carry_deliveries,
+    do_you_have_one_of_the_listed_medical_conditions,
+):
+    result = execute_sql(
+        sql="CALL cv_staging.create_web_submission("
+        ":nhs_number,"
```
thomasridd, line 118: Handle the potential errors on `execute_sql` looks important here

I expect you should probably think about handling all the boto3 errors but here seems particularly important to avoid data loss

thomasridd (changes requested):

vulnerable_people_form/integrations/persistence.py
samcrang: What happens if the `CALL` returns a message/error state indicating that the query failed validation or something? Will that also be handled?

vulnerable_people_form/integrations/persistence.py
danwilliams: This is now done 🙂

vulnerable_people_form/integrations/persistence.py
danwilliams: That will bubble up to Flask, and the user's POST request will be redirected to an error page. The per-page form validation should make it impossible to have invalid data when submitting the 'whole' form, so I think this behaviour is legitimate. In addition, the transaction will exit with a rollback.

pipwilson: @danwilliams we don't currently have a Sentry instance provisioned since we can't use the instance used by GOV.UK which the Rails app was using. I am trying to get a new instance approved and set up at the moment, but we can't currently rely on Sentry being in place.

sanisc: sentry has been removed from the app in the concourse pipeline branch as @pipwilson has pointed out that there is no instance available to use

danwilliams: Correct, as per my earlier answers this behaviour is not dependent upon Sentry - it seems we have removed that elsewhere, which is fine, but out of scope of these changes 🙂

vulnerable_people_form/integrations/postcode_eligibility.py
```diff
@@ -1,6 +1,15 @@
 import os
 
+from .persistence import get_rds_data_client, generate_string_parameter, execute_sql
+
 
 def check_postcode(postcode):
-    # TODO: add implementation
-    return os.environ.get("FAIL_POSTCODE_VERIFICATION") is None
+    records = execute_sql(
+        "CALL cv_base.is_postcode_on_lockdown(:postcode)",
```
sanisc, line 8: The version that looks like it is going through a PR/draft is "CALL cv_base.is_postcode_in_lockdown(:postcode)" so i am guessing this will change to that ?

vulnerable_people_form/integrations/persistence.py
thomasridd: If we go with Flask as-is what will be catching doing our logging?
Just a guess but I'd think with the PaaS out of the box the answer is no
I think that question needs answering but outside of this PR

vulnerable_people_form/integrations/postcode_eligibility.py
danwilliams: Whoops, that got reverted in a rebase. Good spot! Fixed.

thomasridd (approved):
Go ahead and merge this if it is in a good place - subject to adding a second story about how our monitoring and alerting works

sanisc (approved):

PR #3, opened 2020-08-25
SPL validation

danwilliams:
**What**
Added the logic to carry out an SPL check and then send a notification to the user based on the data entered in the form.

**Why**
Users need to know if they are on the SPL and what the next steps are once they have submitted the form.

**How**

Added in all templates for match and no match from SPL
Updated GOV.UK Notify python client to send emails, SMS and letters
Added SPL sproc call to determine whether use is on SPL or not
Updated readme to fix typos

### DIFF
--- a/vulnerable_people_form/form_pages/check_your_answers.py
+++ b/vulnerable_people_form/form_pages/check_your_answers.py
@@ -1,34 +1,13 @@
-from flask import redirect, session
+from flask import session
 
-from ..integrations import govuk_notify_client
 from .blueprint import form
 from .shared.render import render_template_with_title
 from .shared.routing import route_to_next_form_page, dynamic_back_url
 from .shared.session import (
-    form_answers,
     persist_answers_from_session,
     get_answer_from_form,
-    get_summary_rows_from_form_answers,
-    request_form,
-    should_contact_gp,
-)
-
-
-def send_sms_and_email_notifications_if_applicable(reference_number):
-    first_name = get_answer_from_form(["name", "first_name"])
-    last_name = get_answer_from_form(["name", "last_name"])
-    phone_number = get_answer_from_form(["contact_details", "phone_number_texts"])
-    email = get_answer_from_form(["contact_details", "email"])
-    contact_gp = should_contact_gp()
-
-    if phone_number:
-        govuk_notify_client.try_send_confirmation_sms(
-            phone_number, first_name, last_name, contact_gp
-        )
-    if email:
-        govuk_notify_client.try_send_confirmation_email(
-            phone_number, first_name, last_name, reference_number, contact_gp
-        )
+    get_summary_rows_from_form_answers)
+from ..integrations import govuk_notify_client, spl_check
 
 
 @form.route("/check-your-answers", methods=["GET"])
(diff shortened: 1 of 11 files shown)


vulnerable_people_form/integrations/govuk_notify_client.py
```diff
@@ -1,110 +1,68 @@
...
+        },
     )
 
 
-def _try_and_report_exception_to_sentry(fn, *args, **kwargs):
-    try:
-        fn(*args, **kwargs)
-    except Exception as e:
-        sentry_sdk.capture_exception(e)
+def try_send_spl_failure_sms(*args, **kwargs):
```
sanisc, line 63: i think there are letters to be sent as well so that will need a method as well

vulnerable_people_form/form_pages/check_your_answers.py
```diff
@@ -14,20 +14,22 @@
...
-    first_name = get_answer_from_form(["name", "first_name"])
-    last_name = get_answer_from_form(["name", "last_name"])
-    phone_number = get_answer_from_form(["contact_details", "phone_number_texts"])
-    email = get_answer_from_form(["contact_details", "email"])
-    contact_gp = should_contact_gp()
-
-    if phone_number:
-        govuk_notify_client.try_send_confirmation_sms(
-            phone_number, first_name, last_name, contact_gp
+def send_spl_failure_email_if_applicable(reference_number):
```
sanisc, line 17: i don't think in the _init.py file there is any env for the template id's - i know its not in this change log but they will need adding as currently in staging it blows up as the template is is never read in as in env var

vulnerable_people_form/integrations/govuk_notify_client.py
```diff
@@ -1,110 +1,68 @@
...
+                    "Contact your GP or hospital clinician as soon as possible so they can put you on the NHS list of people who should be shielding. You may not get the support you need if you do not contact them.",
+                    "We’ll check your details, then contact you to confirm whether you’re eligible for support. This can take up to 2 weeks from when your GP or hospital clinician puts you on the NHS list of people who should be shielding.",
+                ),
+            ),
+            "Contact your local authority if you need support urgently and cannot rely on family, friends or neighbours: www.gov.uk/coronavirus-local-help",
+        )
+    )
     get_notifications_client().send_sms_notification(
         phone_number=phone_number,
         template_id=current_app.config["GOVUK_NOTIFY_SMS_TEMPLATE_ID"],
```
sanisc, line 39: i will provide new template id's for this there will be 6 in total 3 for each type sms, email and letter and whether there is a match or not

vulnerable_people_form/form_pages/check_your_answers.py
sanisc: i have added all the env vars into master now so if you do a rebase then you should get all the template ids @danwilliams

vulnerable_people_form/integrations/govuk_notify_client.py
sanisc: i have added all the env vars into master now so if you do a rebase then you should get all the template ids @danwilliams

sanisc (approved):

PR #4, opened 2020-08-27
Implement new journey changes

danwilliams:
## Summary of changes

This branch makes a large number of small changes to the journey, in terms of adding/removing pages, adjusting the order of pages, and altering the content of pages, in order to align with the latest updates to the journey spec.

**Note: It depends upon https://github.com/alphagov/covid-engineering/pull/710.**

## How to test

The journey should be run end-to-end.


vulnerable_people_form/form_pages/check_your_answers.py
```diff
@@ -36,14 +36,16 @@ def get_check_your_answers():
...
+        previous_path=dynamic_back_url(),
         summary_rows=get_summary_rows_from_form_answers(),
     )
 
 
 @form.route("/check-your-answers", methods=["POST"])
 def post_check_your_answers():
-    reference_number = persist_answers_from_session()
-    send_sms_and_email_notifications_if_applicable(reference_number)
+    session["nhs_sub"] = "a" * 36
```
sanisc, line 46: whats the reason for creating an nhs sub as 

"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"

vulnerable_people_form/form_pages/nhs_login_callback.py
```diff
@@ -53,4 +53,5 @@ def get_nhs_login_callback():
 
 @form.route("/start", methods=["GET"])
 def get_start():
-    return redirect("/view-or-setup")
+    session.clear()
```
sanisc, line 56: what happens when you go to the default route? As currently it goes to an error page. 

https://gds-shielded-vulnerable-people-service-staging.london.cloudapps.digital/

sanisc, line 56: should we be clearing the session at the start of the journey is there a problem if a user goes all the way back to this page and has already got some info in the session?

vulnerable_people_form/form_pages/shared/session.py
```diff
@@ -225,66 +225,62 @@ def load_answers_into_session_if_available():
...
-            },
-            "contact_details": {
-                "phone_number_calls": contact_number_calls["stringValue"],
-                "phone_number_texts": contact_number_texts["stringValue"],
-                "email": contact_email["stringValue"],
-            },
+            "support_address": _strip_keys_with_no_value(
+                {
+                    "building_and_street_line_1": address_line1["stringValue"],
+                    "building_and_street_line_2": address_line2.get("stringValue"),
```
sanisc, line 251: good to see this is a .get now as before it was causing the issue in staging

sanisc (changes requested):

vulnerable_people_form/form_pages/nhs_login_callback.py
danwilliams: It goes to a 'Page not found' page (404). We can handle this differently if required, I've not received any guidance on what the behaviour should be in this case.

danwilliams: The `/start` route redirects the browser with a 302, consequently `/start` isn't placed in the browser's history, so the back button isn't an issue here.

The `/start` route is mostly there to force a new session to be instantiated, and (in my mind) was intended as the URL that the external service would link to, when a user follows a "start the form" link.

The presence and behaviour of this endpoint is something carried over from the legacy Ruby application. It can (and maybe should) be removed. However, that's likely out of scope of this PR.

sanisc: would it not go to the start of the journey ?

danwilliams: I think it should probably go to the external page.

vulnerable_people_form/form_pages/check_your_answers.py
danwilliams: Good spot - test code, since removed! 🙂

vulnerable_people_form/form_pages/nhs_login_callback.py
sanisc: I have spoken to Mike about this and it should:

'divert to first page of our service 'are you using this service for yourself or for someone else' page'

danwilliams: Updated 🙂 Thanks!

sanisc (approved):

PR #5, opened 2020-08-28
Updated pipeline smoke test URL

sanisc:
Updated the pipeline smoke test URL for staging and prod as it is no longer the link that is currently in the .yml file and throws an error in concourse.


jk132 (approved):
Fine :)

PR #11, opened 2020-09-01
Add unit tests for core business logic

tal-nagra:
What? Initial unit test have been added.
Why? No unit tests were previously present.
How? pytest unit tests have been added as per other COVID engineering repos.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ __pycache__/
 
 /instance/config.py
 /vulnerable_people_form/static
+.coverage
 
 /venv
 .idea
--- a/Makefile
+++ b/Makefile
@@ -4,10 +4,9 @@ test:
 
 install:
 	set -e
-	echo "⏳ installing..."
-	pip3 -q install pytest
-	pip3 -q install -r requirements.txt
-	echo "✔️ Pip dependencies installed!"
+	@echo "⏳ Installing dependencies..."
+	pip3 -q install -r requirements-dev.txt
+	@echo "✔️ Pip dependencies installed!"
 
 concourse_e2e:
 	echo "TODO implement and run behave tests"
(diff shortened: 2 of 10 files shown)


danwilliams (commented):
Some of the formatting is a bit dodgy, and inconsistent, but I'm not sure if that matters at this point or not - we never did get the flake8 config from GDS, so it's hard to say what's right and wrong in all areas.

This is a great start to the tests - I just wonder about the formatting, hence this is a comment and not an approval nor rejection at this stage.

I wonder if the PR description could include any coverage stats? That might be useful...?

sanisc:
> Some of the formatting is a bit dodgy, and inconsistent, but I'm not sure if that matters at this point or not - we never did get the flake8 config from GDS, so it's hard to say what's right and wrong in all areas (although, as a note, our original instruction was to use black - that has been used in the rest of the project, so could be a benchmark to use here for consistency).
> 
> This is a great start to the tests - I just wonder about the formatting, hence this is a comment and not an approval nor rejection at this stage.
> 
> I wonder if the PR description could include any coverage stats? That might be useful...?

The flake8 config will fix any inconsistencies in the app e.g various places where imports aren't used - makes sense to try get this in then run the flake8 stuff to get a baseline

pipwilson:
The flake8 config from either https://github.com/alphagov/digitalmarketplace-api/blob/master/.flake8 or https://github.com/alphagov/covid-engineering/blob/master/.flake8 are OK for you to use.

sanisc:
flask8 has been added as a github action now

sanisc:
> Some of the formatting is a bit dodgy, and inconsistent, but I'm not sure if that matters at this point or not - we never did get the flake8 config from GDS, so it's hard to say what's right and wrong in all areas (although, as a note, our original instruction was to use black - that has been used in the rest of the project, so could be a benchmark to use here for consistency).
> 
> This is a great start to the tests - I just wonder about the formatting, hence this is a comment and not an approval nor rejection at this stage.
> 
> I wonder if the PR description could include any coverage stats? That might be useful...?

This has all been completed so its ready to be merged now.

sanisc (approved):

PR #17, opened 2020-09-03
Add and apply flake8 config

danwilliams:
Correctly added and applied Flake8 config, according to guidelines given by GDS.


sanisc:
Spoken to  theTech Leads in the engineering team and the app should just use flake8 and not black.

danwilliams:
As a note, when the project started, the direction received was to use Black (as per @pipwilson's direction). We then subsequently received indication to use Flake8. This PR applies Flake8, and so satisfies that remit. Removing Black will not really achieve anything given that the project was originally developed using it as a standard - ideally it would continue in the same direction to ensure consistency.

tal-nagra (approved):

samcrang (approved):

PR #28, opened 2020-09-17
Continuously run a smoke test against the form

samcrang:
At the moment this is a very simple smoke test. We should be do
something smarter but this is a start.

If the smoke test is successful then we curl Cronitor which is
configured as a heartbeat to expect a request every _n_ minutes.

PagerDuty can then be triggered off this.

### DIFF
--- a/concourse/pipeline.yml
+++ b/concourse/pipeline.yml
@@ -19,10 +19,10 @@ resources:
       url: ((slack_hook_automate))
       channel: covid-engineering-team
 
-  - name: every-1m
+  - name: every-10m
     type: time
     source:
-      interval: 1m
+      interval: 10m
 
 # Anchors for later
 slack_alert_on_failure: &slack_alert_on_failure
@@ -162,7 +162,7 @@ jobs:
 
   - name: continuous-smoke-test-staging
     plan:
-      - get: every-1m
+      - get: every-10m
         trigger: true
       - get: git-master
       - task: smoke-test
@@ -179,7 +179,7 @@ jobs:
 
   - name: continuous-smoke-test-prod
     plan:
-      - get: every-1m
+      - get: every-10m
         trigger: true
       - get: git-master
       - task: smoke-test
--- a/concourse/pipeline.yml
+++ b/concourse/pipeline.yml
@@ -168,7 +168,7 @@ jobs:
       - task: smoke-test
         file: git-master/concourse/tasks/smoke-test.yml
         params:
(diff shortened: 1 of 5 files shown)


concourse/tasks/smoke-test.yml
```diff
@@ -2,18 +2,23 @@ platform: linux
...
+    - -c
     - |
-      set -eu
-
-      echo "$MESSAGE"
-
-      # TODO we can come up with a more thorough test than this
-      curl --fail --silent --verbose "$URL"
+      echo "running smoke tests against \"${WEB_APP_BASE_URL}\""
+      behave behave/features/ --tags='@e2e_happy_path_no_nhs_login' --stop
```
sanisc, line 23: there is a make command in the make file for this already ?

concourse/pipeline.yml
```diff
@@ -151,6 +156,40 @@ jobs:
...
+      - task: smoke-test
+        file: git-master/concourse/tasks/smoke-test.yml
+        params:
+          WEB_APP_BASE_URL: 'https://coronavirus-shielding-support.service.gov.uk/applying-on-own-behalf'
+          MESSAGE: "Checks that the application deployed to production is not serving HTTP error codes"
+        on_failure: *slack_alert_on_failure
+      - task: cronitor-heartbeat
+        timeout: 1m
+        params:
+          CRONITOR_URL: ((svp-form/cronitor-heartbeat-prod))
```
sanisc, line 194: guess this variable has been added to concourse?

samcrang: Yup, I've got a cronitor account and the URL you have to ping looks like https://cronitor.link/some-secret-key so i wanted to make sure it wasn't public.

concourse/tasks/smoke-test.yml
samcrang: Yeah, I had deliberately duplicated it here.

The Makefile target is called `concourse_e2e`—I've assumed that at some point the set of e2e tests will grow to cover more scenarios. I think for our smoke test it makes sense to just test one(?) happy path (so that we know that the app and DB are working).

I could maybe move this to a Makefile target called `smoke_test` or something?

sanisc: sounds good if you could just move it to a new target like you suggested

concourse/pipeline.yml
sanisc: fantastic

sanisc (approved):

PR #32, opened 2020-09-18
Bug fixes from acceptance testing

sanisc:
Bug fixes that came out of acceptance testing.

### DIFF
--- a/vulnerable_people_form/form_pages/shared/answers_enums.py
+++ b/vulnerable_people_form/form_pages/shared/answers_enums.py
@@ -22,8 +22,8 @@ class ViewOrSetupAnswers(EnumWithText):
 
 @enum.unique
 class ApplyingOnOwnBehalfAnswers(EnumWithText):
-    YES = 1
-    NO = 0
+    YES = 0
+    NO = 1
 
     __text_values__ = {
         YES: "Myself",
--- a/vulnerable_people_form/form_pages/support_address.py
+++ b/vulnerable_people_form/form_pages/support_address.py
@@ -10,8 +10,9 @@ from .shared.validation import validate_support_address
 
 @form.route("/support-address", methods=["POST"])
 def post_support_address():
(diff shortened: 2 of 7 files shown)


vulnerable_people_form/form_pages/support_address.py
```diff
@@ -10,8 +10,9 @@
 
 @form.route("/support-address", methods=["POST"])
 def post_support_address():
-    original_address = {k: v for k, v in form_answers()["support_address"].items() if k != "uprn"}
+    original_address = {k: v for k, v in form_answers().get("support_address", {}).items() if k != "uprn"}
```
tal-nagra, line 13: it might be best to declare get the support address first and check for it's existence before accessing items()

vulnerable_people_form/form_pages/shared/answers_enums.py
```diff
@@ -22,8 +22,8 @@ class ViewOrSetupAnswers(EnumWithText):
 
 @enum.unique
 class ApplyingOnOwnBehalfAnswers(EnumWithText):
-    YES = 1
-    NO = 0
+    YES = 0
```
tal-nagra, line 25: is there any reason these have been swapped around?

tal-nagra (changes requested):

vulnerable_people_form/form_pages/support_address.py
sanisc: it defaults to an empty object if there is no support address so items will always work

vulnerable_people_form/form_pages/shared/answers_enums.py
tal-nagra: Ignore this comment, i have now seen the slack conversation confirming that the underlying enum value needs to be sapped arround

tal-nagra (approved):

PR #38, opened 2020-09-22
Add LICENCE file containing MIT licence and copyright notice

stephengrier:
Add a LICENCE file containing our standard MIT Licence and Crown
Copyright notice.

This is a public repo so it is important we get this added. The GDS Way guidance on licensing is [here](https://gds-way.cloudapps.digital/manuals/licensing.html).


pipwilson (approved):
LGTM.

PR #41, opened 2020-09-24
Update matching messages

StephenGill:
This PR:

- updates content and if statements so that message about contacting your GP appears first where appropriate
- adds sentence covering use case where user fails to match on date of birth
- updates links to point to correct url - gov.uk/coronavirus-shielding-support
- corrects minor inconsistencies with formatting

Trello card: https://trello.com/c/bIvyvj2v/335-update-notify-content-to-reflect-matching-behaviour


sanisc:
To avoid a commit from a fork this has been replicated internally by a dev

PR #48, opened 2020-09-25
Exclude non-postable addresses from address picker

stephengrier:
The address picker gets address data for postcodes from the Ordnance
Survey Places API. The API supports two datasets, a DPA dataset, which
is compiled by Royal Mail and includes all residential and most
commercial postal addresses; and an LPI dataset, which is compiled by
local authorities and contains both postal addresses, and non-postable
assets such as carparks and streets, which should not appear in the
address picker. We are using the LPI dataset on the address picker.

The address picker should probably use the DBA dataset. However, DPA
records have a different structure to LPI records, which would require a
more extensive rewrite of the postcode_lookup_helper. The documentation
about these datasets is poor, so I've made the decision to simply
exclude records where `POSTAL_ADDRESS_CODE=N`, which indicates a
non-postable address.

This probably needs a more thorough rewrite at some point to make sure
we are using the OS Places API efficiently, and to add a thorough set of
unit tests.

Also, add some pytest tests for the postcode_lookup_helper module.

1. https://apidocs.os.uk/docs/os-places-dpa-output
2. https://apidocs.os.uk/docs/os-places-lpi-output
3. https://gbgplc.zendesk.com/hc/en-us/articles/207710315-AddressBase-Premium-Codes

### DIFF
--- a/tests/integrations/test_postcode_lookup_helper.py
+++ b/tests/integrations/test_postcode_lookup_helper.py
@@ -1,0 +1,100 @@
+import pytest
+import requests
+import json
+from unittest import mock
+from flask import Flask
+
+from vulnerable_people_form.integrations.postcode_lookup_helper import \
+        entry_is_a_postal_address, \
+        get_addresses_from_postcode, \
+        PostcodeNotFound, ErrorFindingAddress
+
+_current_app = Flask(__name__)
+
+
+def test_entry_is_a_postal_address_function(fake_os_places_api_entry):
+    postable_entry = fake_os_places_api_entry(postal_address_code='D')
+    non_postable_entry = fake_os_places_api_entry(postal_address_code='N')
+
+    assert entry_is_a_postal_address(postable_entry.to_json()) is True
+    assert entry_is_a_postal_address(non_postable_entry.to_json()) is False
+
+
+@mock.patch('requests.get')
+def test_get_addresses_from_postcode(mock_get, faker, fake_os_places_api_entry, fake_os_places_api_response):
+    postcode = faker.postcode()
+
+    entries = []
+    # GIVEN 5 postable address entries for a postcode
+    for _ in range(0, 5):
+        entries.append(fake_os_places_api_entry(postcode=postcode, postal_address_code='D'))
+    # and one non-postable entry.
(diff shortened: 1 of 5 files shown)


tests/integrations/test_postcode_lookup_helper.py
```diff
@@ -0,0 +1,100 @@
...
+def test_get_addresses_from_postcode_handles_500_status_code(mock_get, faker):
+    postcode = faker.postcode()
+    mock_get.return_value = create_response_object(500, '')
+    with pytest.raises(ErrorFindingAddress):
+        with _current_app.test_request_context():
+            get_addresses_from_postcode(postcode)
+
+
+def api_response_and_function_result_should_match(api_response, addresses):
+    for i in range(0, len(api_response.address_entries)):
```
VivianAllen, line 77: I think it might be better to filter out the unpostable addresses before calling this function, rather than having a switch that ignores them? Maybe it doesn't make much difference, but it feels like you're relying on the indexes still matching after you skip over 'N' entries and other things, seems like it could be brittle if the tests had to be extended much.

VivianAllen (commented):
looks really good. A small question about how 'N' postcodes are handled in the test asserts, which I think is worth discussing, but otherwise this looks solid to me.

tests/integrations/test_postcode_lookup_helper.py
sanisc: I agree, i think it can be merged without this but would be something to look into if there is time

sanisc (dismissed):

tests/integrations/test_postcode_lookup_helper.py
stephengrier: I've changed the test to exclude the non-postable entry from the list of expected entries passed to the compare function. It also now asserts the number of addresses returned is equal to the number of expected entries. Is this better?

VivianAllen: works for me! 👍

VivianAllen (approved):
looks good to me!

sanisc:
LGTM @stephengrier

sanisc (approved):

PR #52, opened 2020-09-28
Add missing word to message templates

StephenGill:
This PR adds a missing word ('need') to the 3 matching message templates.

@sanisc


sanisc:
created again not as a fork but as a commit to the repo

PR #78, opened 2020-10-15
Add missing config values to Config.scala for SvpSimulationConstantUs…

VivianAllen:
…ersPerSec

Several missing config variables were preventing SvpSimulationConstantUsersPerSec
from building.


loadtests/src/test/main/scala/svp/Config.scala
```diff
@@ -13,6 +13,9 @@ object Config {
   val injectDurationSeconds: Int = System.getProperty("injectDurationSeconds", "10").toInt
   val pauseMin: FiniteDuration = getProperty("pauseBetweenRequestsInSecondsMin", "3").toInt.seconds
   val pauseMax: FiniteDuration = getProperty("pauseBetweenRequestsInSecondsMax", "6").toInt.seconds
+  val responseTimeMs = 800
```
sanisc, line 16: this already exists above in this file

sanisc (changes requested):

loadtests/src/test/main/scala/svp/Config.scala
VivianAllen: fixed

sanisc (approved):

PR #79, opened 2020-10-15
PoC: Working gatling tests.

whpearson:
I moved the scala directory back to not being beneath main (as that
seems to be the standard dir).

I found that two variables seem to be renamed, so want to verify their
use before fully changing things.

This starts to run the test when run locally in the same mvn container.


whpearson:
I've resolved the variables to be the same as already merged on master.

sanisc (approved):

PR #80, opened 2020-10-15
Add PROD environment (automated) smoke test

tal-nagra:
A smoke test for the production environment is required
that will not record a submission to the PROD database.

A new behave feature file has been added for the happy
path journey that progresses to the /check-your-answers
page (just prior to persisting answers to the database).

The Makefile and the concourse smoke-test.yml have been
updated accordingly.

### DIFF
--- a/behave/features/environment.py
+++ b/behave/features/environment.py
@@ -8,13 +8,15 @@ from selenium.webdriver.chrome.options import Options
 def before_all(context):
     chrome_options = Options()
     chrome_options.binary = "bin/headless-chromium"
-    chrome_options.add_argument("-headless")
-    chrome_options.add_argument("--no-sandbox")
+    chrome_options.headless = True
+    chrome_options.add_argument("--no-sandbox")  # Bypass OS security model
     chrome_options.add_argument("--single-process")
     chrome_options.add_argument("--disable-dev-shm-usage")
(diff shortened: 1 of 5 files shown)


behave/features/environment.py
```diff
@@ -8,13 +8,15 @@
...
+    chrome_options.headless = True
+    chrome_options.add_argument("--no-sandbox")  # Bypass OS security model
     chrome_options.add_argument("--single-process")
     chrome_options.add_argument("--disable-dev-shm-usage")
+    chrome_options.add_argument("window-size=1200,800")
 
     context.browser = webdriver.Chrome(options=chrome_options)
-    context.browser.set_page_load_timeout(time_to_wait=200)
+    context.browser.set_page_load_timeout(time_to_wait=30)
+    context.browser.delete_all_cookies()
```
thomasridd, line 19: did wonder about that one

thomasridd: (not in a bad way but whether they were persisting)

thomasridd (approved):
each bit looks good - let's see what it does

behave/features/environment.py
tal-nagra: hopefully any cookie related issues will be eradicated (if they exist that is)

PR #82, opened 2020-10-15
Updating gatling data with postcodes

whpearson:
These postcodes should match what is in the staging environment

I mapped all the EX postcodes to the same one, but otherwise gave
a unique postcode to test variety of scenarios.


whpearson:
A good way to check all the postcodes is 
`cat loadtests/src/test/resources/data/svp_submission_data.json | grep postcode | sort | uniq -c`
And then check that they are all in the from the right file.

sanisc (approved):

PR #85, opened 2020-10-15
Update gatling tests for MI

whpearson:
We have a better idea of what are MI pipeline looks like now.

It will run right after our main pipeline so we no longer need
to trigger a job (and that job was not the correct one).

Also the total pipeline will take longer, so load test for
30 mins instead of 20.


VivianAllen (approved):
oof, theres a lot to unpack here, but I think everything looks right!  ✅

sanisc (approved):

PR #89, opened 2020-10-16
847 Correct Unused postcodes

whpearson:
Two of the postcodes in the previous commit turned out to be deprecated.
While they were valid postcodes it changed the flow of the form, which
broke the gatling tests.

Both new postcodes are in use and valid postcodes for registration.


junaidsharif (approved):
LGTM 👍

jk132 (approved):
Looks sensible, changes as per group discussion.

PR #91, opened 2020-10-16
847/scale web form in response to failing load tests

jk132:
Change CF settings to increase memory and instances of app

Given failing Gatling load tests, increase instances to 4 (from 1) and memory
from 1G to 2G to test improvement on performance


VivianAllen (approved):
seems sensible ✅

PR #92, opened 2020-10-16
Revert "847/scale web form in response to failing load tests"

jk132:
Reverts alphagov/govuk-shielded-vulnerable-people-service#91
Turns out these instance numbers were being set in the Concourse task (overriding those set in this manifest), so it tried to create 50 instances each with 2GB memory, which exceeded our PaaS quota. Reverting this change.


whpearson (approved):
Looks good.

PR #93, opened 2020-10-16
847 Match number of prod instances in staging

whpearson:
Staging was struggling to serve gatling requests. This PR matches
staging to prod, both with 30 instances.


jk132 (approved):
LGTM

PR #94, opened 2020-10-16
Lower prod instances temporarily for gatling staging test

jk132:
We want to test app in staging with same number of instances as well have for
prod, need to decrease prod number of instances temporarily so we can do this
without exceeding quota


whpearson (approved):
LGTM

PR #95, opened 2020-10-16
847 Investigation of connection sharing

whpearson:
We noticed lots of connections to the database and that we were creating
a new client each time we did a query.

This is an investigation into maintaining a persistent http connection
to aws.

It uses the flask g object to store a connection which has a the headers
set to keep alive.

### DIFF
--- a/vulnerable_people_form/integrations/persistence.py
+++ b/vulnerable_people_form/integrations/persistence.py
@@ -4,6 +4,7 @@ import boto3
 import botocore.exceptions
 from botocore.config import Config
 from flask import current_app
+from flask import g
 
 boto3_config = Config(
     retries={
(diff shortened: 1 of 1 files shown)


sanisc (commented):
what happens when the app is torn down does the connection still exist ?

whpearson:
Ah, good point. I'll add some cleanup.

vulnerable_people_form/integrations/persistence.py
```diff
@@ -22,8 +23,16 @@ def get_client_kwargs(app=current_app):
     }
 
 
+def set_connection_header(request, operation_name, **kwargs):
+    request.headers['Connection'] = 'keep-alive'
+
+
 def get_rds_data_client(app=current_app):
-    return boto3.client("rds-data", **get_client_kwargs(app))
+    if 'boto_rds_client' not in g:
```
VivianAllen, line 31: this is really neat! I presume its more reliable than just instantiating a global client on import?

whpearson: Just going by documentation on how to maintain db connections. https://flask.palletsprojects.com/en/1.1.x/appcontext/

You could have some retry logic here if the client gets in a bad state.

VivianAllen: cool stuff!

VivianAllen (approved):
lgtm ✅

whpearson:
Going to hold off merging, until we have a run at the current level.

vulnerable_people_form/integrations/persistence.py
sanisc: what happens if it does get a bad state and the app is still running ? Is that something we can pre-empt?

whpearson:
Closing for now, as performance seems okay for our needs.

PR #96, opened 2020-10-16
Revert temp staging-prod instance count switch

jk132:
After running gatling on staging with 30 instances of the app, revert the
config to its stable version prior to Monday go-live. Further investigation
will be done to find long-term solution to failing gatling tests.


whpearson (approved):
Looks good

PR #97, opened 2020-10-19
847 Reduce load to around 100 page requests per second

whpearson:
We need to simulate 100 user simulataenous users (not total web
requests). This change tries to hit this requirement rather than trying
to hit 1000 requests per second (as this is hard to tell from gatling,
and can cause confusion).

Also make the pause between request time (1 seconds) be picked up by the tests.


VivianAllen (approved):
lgtm ✅

PR #100, opened 2020-10-19
847 Temporary increase in gatling users

whpearson:
In order to see how we might do in extreme peaks of users
increase number of users to 7, which should be about 210 users
in flight at a time.


VivianAllen (approved):
✅

PR #102, opened 2020-10-19
Revert "847 Temporary increase in gatling users"

whpearson:
Reverts alphagov/govuk-shielded-vulnerable-people-service#100

As we don't want to run with extra users over the night.


VivianAllen (approved):
👍

PR #103, opened 2020-10-19
847 Make gatling tests save a zip file

whpearson:
To make it easier to download a gatling run, zip up the gatling test
output before uploading to s3


VivianAllen (approved):
nice! ✅

PR #114, opened 2020-10-21
867: Use gds metrics library

whpearson:
Gds has a metrics library for python and prometheus. To make sure that
we are sending metrics in the same way, use the library.

https://github.com/alphagov/gds_metrics_python

In order to get metrics working properly you need to let gunicorn know
when a worker has died.

See https://github.com/prometheus/client_python#multiprocess-mode-gunicorn
for more details.

The other options are the same that notify-api uses, which may help with
performance.

https://github.com/alphagov/notifications-api/blob/1d3f9589eae16bd015bf6ed62200aa6d1a60f3a5/gunicorn_config.py


gunicorn_config.py
```diff
@@ -0,0 +1,5 @@
+
+from gds_metrics.gunicorn import child_exit  # noqa
+
+worker_class = "eventlet"
+worker_connections = 256
```
sanisc, line 5: How does this affect performance as the there are 6 workers per paas instance?

whpearson: So with the eventlet worker_class it will do non-blocking io. So when it is talking to the database it will still serve other requests. The worker_connections limits the number of connections that each worker can do in this manner.

I can remove if you think there is something that this would interfere with.

whpearson: It is used on the paas already for notify. 

Something that walks through it's performance implications https://medium.com/@genchilu/brief-introduction-about-the-types-of-worker-in-gunicorn-and-respective-suitable-scenario-67b0c0e7bd62

whpearson: I've removed the worker_class eventlet as it was causing problems with the oidc client.

whpearson:
I've also added some config to help spin up local dev environments. Which needs a bit of thought.

whpearson:
There is some changes to be made to the alembic files (perhaps adding in a new one to target port 3307), if we stick with separate dbs for alembic and frontend dev.

Procfile
```diff
@@ -1 +1 @@
-web: gunicorn -b 0.0.0.0:$PORT -w $GUNICORN_WORKER_COUNT "vulnerable_people_form:create_app(None)"
+web: gunicorn -b 0.0.0.0:$PORT -w $GUNICORN_WORKER_COUNT "vulnerable_people_form:create_app(None)" -c gunicorn_config.py
```
sanisc, line 1: is this gunicorn_config.py still relevant ? As the file has a single import just referencing the child_exit function

whpearson: I think so. That is part of the setup for the metrics library. 

https://github.com/alphagov/gds_metrics_python

README.md
```diff
@@ -91,22 +91,32 @@ To set up the app for local development you need to follow these steps:
     ```sh
     cp instance/config.py.sample.dev instance/config.py
     ```
+2. Start rds-client-api and mysql 
+    ```docker-compose up -d```
 
-2.  Fill in the missing configuration values as per the **Configuration
+
+3. Make sure the database is up to date.
+   ```
```
tal-nagra, line 99: It would be worth mentioning here that a relevant venv should be created and the pip packages installed

requirements.txt
```diff
@@ -5,11 +5,13 @@ govuk-frontend-jinja==0.2.1
...
 python-stdnum==1.13
 oic==1.2.0
 blinker==1.4
 gunicorn==20.0.4
 sentry-sdk==0.17.3
 flask-talisman == 0.7.0
 bleach == 3.2.1
+prometheus-client==0.8.0
+gds-metrics==0.2.4
+
```
tal-nagra, line 17: blank line can be removed

Dockerfile
```diff
@@ -1,13 +1,16 @@
-from python:3.8-alpine
+from python:3.8-slim-buster
```
tal-nagra, line 1: perhaps update this to 3.8.5-slim-buster

tal-nagra (changes requested):

Dockerfile
whpearson: 3.8 references the latest support 3.8 version python. In this case

root@9dd4677296c3:/app# python --version
Python 3.8.6

tal-nagra (approved):
This pull request includes some good work to help get the app running locally end-to-end.

Some further work is needed (outside of this PR) to get the happy path user journey working locally in it's entirety.

PR #115, opened 2020-10-22
847: Remove recursive for s3 upload and fix date

whpearson:
The date could change between invocations, fix this date so it is a
constant.

We are also only uploading a single file, so recursive isn't needed.


VivianAllen (approved):
makes sense

PR #118, opened 2020-10-23
Update link on nhs login landing page (sign out)

sanisc:
The link on the signed out page via nhs login should point to the start of the journey for the user which is the gov uk page


VivianAllen (approved):
seems simple enough

PR #120, opened 2020-10-26
Update NHS login no consent given to log warning

sanisc:
If a user decides to use NHS login to register with the service they must accept and give consent to share their information with the SVP app. Currently if they reject consent it is logged as an error this has been changed to a warning.


vulnerable_people_form/form_pages/nhs_registration_callback.py
```diff
@@ -14,11 +13,14 @@
     set_form_answers_from_nhs_user_info,
     load_answers_into_session_if_available
 )
+from ..integrations import google_analytics
+
+CONFIRMATION_URL = "/confirmation"
```
tal-nagra, line 18: as CONFIRMATION_URL isn't intended to be used outside this module it can be prefixed with underscore

tal-nagra (changes requested):

tal-nagra (approved):

PR #128, opened 2020-10-29
867 Add false value for METRICS_BASIC_AUTH

whpearson:
Else if defaults to true.

We have this restricted to the VPN, so it doesn't need basic auth, and makes testing metrics more complicated.


concourse/tasks/deploy-to-govuk-paas.yml
```diff
@@ -24,6 +24,7 @@ params:
   GOVUK_NOTIFY_NO_SPL_MATCH_EMAIL_TEMPLATE_ID:
   GOVUK_NOTIFY_NO_SPL_MATCH_SMS_TEMPLATE_ID:
   GOVUK_NOTIFY_NO_SPL_MATCH_LETTER_TEMPLATE_ID:
+  METRICS_BASIC_AUTH: false
```
sanisc, line 27: it also needs to be set further down in the actual environment that paas will run

sanisc (changes requested):

sanisc (approved):

PR #129, opened 2020-10-30
Update postcode to an invalid one

junaidsharif:
Smoke test uses postcode which should never be in lockdown. Hence changing Leeds postcode to Edinburgh one. Thus this test will always be successful on account of postcode (since we wont have Edinburgh postcode in lockdown list).

### DIFF
--- a/behave/features/5.not_eligible_postcode.feature
+++ b/behave/features/5.not_eligible_postcode.feature
@@ -18,6 +18,6 @@ Feature: COVID-19 Shielded vulnerable people service - partial user journey - in
 
     Scenario: Should be re-directed to not eligible postcode when unsupported postcode entered
         Given I am on the "postcode-eligibility" page
-        When I give the "#postcode" field the value "LS1 1BA"
+        When I give the "#postcode" field the value "QI1 2ZZ"
         And I submit the form
         Then I am redirected to the "not-eligible-postcode" page


sanisc (commented):
the other tests use BB1 1TA which is a real postcode is it okay to keep using that  ?

jk132:
> the other tests use BB1 1TA which is a real postcode is it okay to keep using that ?

Good point. Where that needs to be in tier 3? Can't remember if we discussed having a postcode in staging purely for testing purposes but using another non-viable (e.g. Scottish) postcode which we add to staging's eligible postcodes list (if required) such that it's always tier 3 could be an option here?

stephengrier:
> > the other tests use BB1 1TA which is a real postcode is it okay to keep using that ?
> 
> Good point. Where that needs to be in tier 3? Can't remember if we discussed having a postcode in staging purely for testing purposes but using another non-viable (e.g. Scottish) postcode which we add to staging's eligible postcodes list (if required) such that it's always tier 3 could be an option here?

Relying on specific postcodes to be in the database is always going to trip us up, particularly in prod. I'd be nervous about using a Scottish postcode even in staging, given that this repo is public.

We discussed changing the frontend to accept a secret HTTP header that makes it treat any postcode as valid, which the smoke test can use to get reliable responses from the frontend. For invalid postcodes we should use a fictitious postcode.

stephengrier (dismissed):

junaidsharif:
> > > the other tests use BB1 1TA which is a real postcode is it okay to keep using that ?
> > 
> > 
> > Good point. Where that needs to be in tier 3? Can't remember if we discussed having a postcode in staging purely for testing purposes but using another non-viable (e.g. Scottish) postcode which we add to staging's eligible postcodes list (if required) such that it's always tier 3 could be an option here?
> 
> Relying on specific postcodes to be in the database is always going to trip us up, particularly in prod. I'd be nervous about using a Scottish postcode even in staging, given that this repo is public.
> 
> We discussed changing the frontend to accept a secret HTTP header that makes it treat any postcode as valid, which the smoke test can use to get reliable responses from the frontend. For invalid postcodes we should use a fictitious postcode.

and just to add on top of whats said before.. non-smoke tests relies on a valid web submission… and they need a postcode which is in lockdown… both of these conditions are met by BBA 1TA. Hence we are good for now. This will of course needs adjustment when we go with solution re secret HTTP header.

stephengrier (approved):
Using a fictitious postcode fixes the `5.not_eligible_postcode.feature` behave journey, the other journeys need addressing but I think this pull request is 👍

PR #139, opened 2020-11-02
Up gatling tests to 200 active users

whpearson:
We wish to stress test the service due to changes in data
in the postcode table.

Temporarily increase the users to 200 per second.


junaidsharif (approved):
LGTM! 👍

PR #140, opened 2020-11-02
Revert "Up gatling tests to 200 active users"

whpearson:
We don't want 200 active users running over night, as they are not our none-functional requirements.

Reverts alphagov/govuk-shielded-vulnerable-people-service#139


junaidsharif:
As we have gone national, shouldn't it be prudent that we keep this as  8 users / 240 requests ? 

This is more than NFR (100 requests) but gives us a bit more sight in case we start going downhill ?
@stephengrier @pipwilson @whpearson

PR #144, opened 2020-11-04
Fix e2e-test-staging for deployment

junaidsharif:
This PR excludes the backend testing in deployment pipeline.

The original PR was tested using a stand-alone pipeline. However, there was a component missed where e2e test executes all features except no submission. This PR adds the backend test also as exclusion to e2e tests done during deployment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,8 +14,9 @@ install:
 
 concourse_e2e:
 	@echo "Executing e2e automated tests against the staging environment..."
-	# execute all scenarios except no submission ( the '-' character prefix means skip)
-	behave behave/features --stop --tags=-e2e_happy_path_no_nhs_login_no_submission
+	# execute all scenarios except no submission and backend submissions test ( the '-' character prefix means skip)
+	behave behave/features --stop --tags=-e2e_happy_path_no_nhs_login_no_submission \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -14,8 +14,9 @@ install:
 
 concourse_e2e:
 	@echo "Executing e2e automated tests against the staging environment..."
-	# execute all scenarios except no submission ( the '-' character prefix means skip)
-	behave behave/features --stop --tags=-e2e_happy_path_no_nhs_login_no_submission
+	# execute all scenarios except no submission and backend submissions test ( the '-' character prefix means skip)
+	behave behave/features --stop --tags="-@e2e_happy_path_no_nhs_login_no_submission and -@simple_web_submission_entry \
```
tal-nagra, line 18: this command is not excluding the `@s3_outputs_check test`
The `behave` tag expressions documentation might be of some use:
https://behave.readthedocs.io/en/latest/behave.html#tag-expression
https://behave.readthedocs.io/en/latest/tag_expressions.html

tal-nagra (changes requested):

tal-nagra (approved):
Going forward we should consider whether we just add a new tag that includes features rather than excludes. However, this can be discussed/tackled when we add the next set of features.

PR #145, opened 2020-11-04
Fix (attempt) concourse failing smoke tests

tal-nagra:
When the concourse pipeline executes the smoke
tests it intermittently fails. This behaviour
has started to manifest after the
gdscyber/concourse-chrome-driver has been updated.

A --disable-gpu chrome driver configuration option
is now supplied in an attempt to remedy the
'Timed out receiving message from renderer' error.


whpearson:
The config works locally. It seems unlikely to fix our problem (as it seems to be hitting smoke tests on two environments at the same time).

whpearson (approved):

PR #152, opened 2020-11-06
Remove back link from nhs registration page

sanisc:
Update the design to match figma as this page is not designed to have a back button


VivianAllen (approved):
lgtm ✅

PR #156, opened 2020-11-06
Remove back button from view answers page

sanisc:
The back link should not be present on the view answers page as this does not align with figma


VivianAllen (approved):
lgtm ✅

PR #165, opened 2020-11-11
Task/update address validation

sanisc:
The address line one and two fields were limited to 75 characters
by the database and input validation. The OS places API returns
addresses with that may contain address line two fields longer than
75 characters.
    
This pull request amends the input validation to permit up to 100
characters for address line 1 and 2. Furthermore, leading and trailing
whitespace is not stripped from the user-entered address field values
and unit tests have been added to cover the support_address input
validation.

### DIFF
--- a/tests/form_pages/shared/test_validation.py
+++ b/tests/form_pages/shared/test_validation.py
@@ -431,6 +431,64 @@ def test_validate_date_of_birth_should_return_false_when_an_invalid_date_of_birt
     )
 
 
+def test_validate_support_address_should_return_true_when_a_valid_address_is_provided():
+    with _current_app.test_request_context() as test_request_ctx:
+        test_request_ctx.session["form_answers"] = {
+            "support_address": {
+                "building_and_street_line_1": "123 Test Avenue",
+                "building_and_street_line_2": "",
+                "town_city": "Leeds",
+                "postcode": "LS11BA"
+            }
+        }
+        is_valid = validation.validate_support_address()
+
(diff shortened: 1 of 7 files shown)


vulnerable_people_form/form_pages/shared/validation.py
```diff
@@ -252,13 +252,13 @@ def validate_support_address():
         [
             validate_length(
                 ("support_address", "building_and_street_line_1"),
-                75,
-                length_fstring.format("Address line 1", 75),
+                100,
+                length_fstring.format("Address line 1", 100),
```
sanisc, line 256: its 110 for the address line one 
and 210 for the address line two

tests/form_pages/shared/test_validation.py
```diff
@@ -431,6 +431,64 @@ def test_validate_date_of_birth_should_return_false_when_an_invalid_date_of_birt
...
+        "town_city": "town / city too long town / city too long town / city too long town / city too long town / city too long ",  # noqa
+        "postcode": "LS11BA"
+     },
+     [{
+         "form_field": "building_and_street_line_1",
+         "expected_error_msg": "'Address line 1' cannot be longer than 100 characters"
+      },
+      {
+         "form_field": "building_and_street_line_2",
+         "expected_error_msg": "'Address line 2' cannot be longer than 100 characters"
```
sanisc, line 472: these need updating with the correct values below

tal-nagra (approved):

PR #166, opened 2020-11-11
Modify E2E test case for back-end validation of supermarket feed

junaidsharif:
This commit makes following changes
- Modify web form submission steps to add date suffix in #first_name field.
- Modify e2e pipeline submission to check supermarket feed in the first name field for yesterday's date
- Retrofit LA feed check to keep code lean.

### DIFF
--- a/behave/features/steps/general.py
+++ b/behave/features/steps/general.py
@@ -4,6 +4,7 @@ from behave import then, when, given
 from selenium.webdriver.common.by import By
 from selenium.webdriver.support import expected_conditions
 from selenium.webdriver.support.wait import WebDriverWait
+from datetime import date
 
 _BASE_URL = os.environ["WEB_APP_BASE_URL"]
 _DEFAULT_TIMEOUT = 30  # in seconds
@@ -54,6 +55,9 @@ def click_element(context, css_selector):
 def set_element_value(context, css_selector, element_value):
     html_element = context.browser.find_element_by_css_selector(css_selector)
     html_element.clear()
+    if css_selector == '#first_name':
+        date_stamp = date.today().strftime('%Y-%m-%d')
+        element_value = element_value+'_'+date_stamp
     html_element.send_keys(element_value)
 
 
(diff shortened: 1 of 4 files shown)


behave/features/steps/general.py
```diff
@@ -54,6 +55,9 @@ def click_element(context, css_selector):
 def set_element_value(context, css_selector, element_value):
     html_element = context.browser.find_element_by_css_selector(css_selector)
     html_element.clear()
+    if css_selector == '#first_name':
+        date_stamp = date.today().strftime('%Y-%m-%d')
+        element_value = element_value+'_'+date_stamp
```
tal-nagra, line 60: It would be cleaner if you create a new function to set the element value to today's date and revert the amends to this function
e.g. `def set_element_value_to_date_today(context, css_selector)`

tal-nagra (changes requested):

behave/features/steps/general.py
```diff
@@ -67,3 +68,12 @@ def set_field_value_step(context, field_name, value):
     elem = context.browser.find_element_by_name(field_name)
     elem.send_keys(value)
     elem.submit()
+
+
+@when('I give the "{css_selector}" field the value "{element_value}" appended with todays date')
+def set_element_value_appended_with_today_date(context, css_selector, element_value):
+    html_element = context.browser.find_element_by_css_selector(css_selector)
```
tal-nagra, line 75: you can just invoke the `set_element_value` function here with your `element_value` variable

this amend will avoid any duplicated code in this pull request

tal-nagra (changes requested):

behave/features/steps/general.py
junaidsharif: addressed as commented !!!!

behave/features/steps/general.py
junaidsharif: and done !

tal-nagra (approved):

PR #167, opened 2020-11-11
Task/881 882 dynamic journey based on tier

whpearson:
This adds all the tiering logic so far. There is a feature flag TIERING_LOGIC, behind which the following functionality is hidden.

* It calls the get_postcode_tier stored procedure in a special blue print for posts with a postcode.
* It calls a separate web submission stored procedure, one that has the tier as well.

There is also some work done on getting end to end running locally. It introduces two environment variables OVERRIDE_ONS_URL and NOTIFY_DISABLED. OVERRIDE_ONS_URL is used to develop locally to hit a very small docker container that acts somewhat like the ONS. Notify is just disabled, as we don't rely on any data from it to continue.

A draft PR. I need to squash some of these commits together, but it won't be easy due to rebases with master. Let me know if further changes are needed before the squash, we also need to merge the database work in and get it pushed out.


whpearson:
get_postcode_tier is stubbed out until we know how to proceed in the back end code implementation

concourse/pipeline.yml
```diff
@@ -76,6 +76,43 @@ jobs:
...
+          HOSTNAME: gds-shielded-vulnerable-people-service-sandbox
+          GOVUK_NOTIFY_SPL_MATCH_EMAIL_TEMPLATE_ID: 80ea4bbd-66f1-455d-b1d2-608e2b8aa948
+          GOVUK_NOTIFY_SPL_MATCH_SMS_TEMPLATE_ID: 17529024-96a3-42f1-945e-9ca50dc87617
+          GOVUK_NOTIFY_SPL_MATCH_LETTER_TEMPLATE_ID: 6a7a0e2c-5e04-47c5-b8ae-686e4a5cfbd2
+          GOVUK_NOTIFY_NO_SPL_MATCH_EMAIL_TEMPLATE_ID: ff27bf85-bddb-451d-9f49-bd862474f29d
+          GOVUK_NOTIFY_NO_SPL_MATCH_SMS_TEMPLATE_ID: b29626c4-7db2-4642-8104-dd01f1b9e663
+          GOVUK_NOTIFY_NO_SPL_MATCH_LETTER_TEMPLATE_ID: a3a3f86a-b2e0-464b-a74d-ec2e7fd29ac5
+          NOTIFY_API_KEY: ((svp-form/notify-api-key-stg))
+          SECRET_KEY: ((svp-form/flask-secret-key-base-stg))
+          NHS_OIDC_AUTHORITY_URL: https://auth.aos.signin.nhs.uk
```
sanisc, line 101: for this to work for nhs login we will need to contact them to add a new address in for the redirect URLs as this will live on a different domain once deployed to staging

whpearson: I'm going to move this to a different PR. But a couple of questions as you've raised this here.

 Is this a blocker for getting it out initially? How can we contact them?

PR #168, opened 2020-11-13
Use new stored procedures with tiering

samcrang:
This change makes the web form able to store tiering information against
a submission but just stores NULL for the time being.

We are also in the process of migrating this new stored procedure to be
the same as the original stored procedure.

### DIFF
--- a/vulnerable_people_form/form_pages/shared/session.py
+++ b/vulnerable_people_form/form_pages/shared/session.py
@@ -213,6 +213,7 @@ def persist_answers_from_session():
         form_answers().get("do_you_have_someone_to_go_shopping_for_you"),
         form_answers().get("medical_conditions"),
         lives_in_england,
+        None,
     )
 
     session["form_uid"] = submission_reference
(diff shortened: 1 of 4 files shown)


vulnerable_people_form/form_pages/shared/session.py
```diff
@@ -293,6 +295,7 @@ def load_answers_into_session_if_available():
             "basic_care_needs": do_you_need_help_meeting_your_basic_care_needs["longValue"],
             "do_you_have_someone_to_go_shopping_for_you": do_you_have_someone_to_go_shopping_for_you["longValue"],
             "do_you_live_in_england": do_you_live_in_england.get("longValue"),
+            "tier_at_submission": tier_at_submission["longValue"],
```
sanisc, line 298: this should be get as the value can be null

sanisc (approved):

vulnerable_people_form/form_pages/shared/session.py
samcrang: Ahhh, you're right

sanisc (approved):

PR #169, opened 2020-11-13
Remove ASDA from supermarket deliveries page

sanisc:
Remove ASDA from the list as Asda have stopped accepting new referrals from the NSSS


junaidsharif (approved):
LGTM 👍  and pretty straight forward !

PR #170, opened 2020-11-14
Modify the SPL check to point to performance efficient temp version

junaidsharif:
This PR modifies the app to use a new efficient temp versio of SPL check sproc. This is needed
so that the original/master sproc is updated without giving Front end app any downtime.

This should be deployed inline with instructions on https://github.com/alphagov/covid-engineering/pull/1001

### DIFF
--- a/vulnerable_people_form/integrations/spl_check.py
+++ b/vulnerable_people_form/integrations/spl_check.py
@@ -8,11 +8,13 @@ init_logger(logger)
 
 
 def check_spl(nhs_number, date_of_birth):
+    test_mode = 'NO'
     records = execute_sql(
-        "CALL cv_base.is_person_currently_on_the_spl(:nhs_number, :date_of_birth)",
+        "CALL cv_base.is_person_currently_on_the_spl_new(:nhs_number, :date_of_birth, :test_mode)",
         (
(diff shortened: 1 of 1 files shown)


vulnerable_people_form/integrations/spl_check.py
```diff
@@ -8,11 +8,13 @@
 
 
 def check_spl(nhs_number, date_of_birth):
+    test_mode = 'NO'
     records = execute_sql(
-        "CALL cv_base.is_person_currently_on_the_spl(:nhs_number, :date_of_birth)",
+        "CALL cv_base.is_person_currently_on_the_spl_new(:nhs_number, :date_of_birth, :test_mode)",
```
sanisc, line 13: what does test_,mode mean?

junaidsharif: Linked with https://github.com/alphagov/covid-engineering/pull/1001

Still under review on db side... This is done because mysql doesnt support optional parameters... and we need to bypass transactions state setting in test mode.... hence the flag which toggles the same in stored procedure..

junaidsharif:
Not needed anymore as original implementation of #1001 doest need this anymore. Closing PR/branch

PR #171, opened 2020-11-14
Modify  SPL Check integration to point to strategic resource

junaidsharif:
This PR modifies the app to use a new efficient version of SPL check sproc. This is needed
so that the original/master sproc is updated without giving Front end app any downtime.

Also, this will allow temp version to be removed.

This should be deployed inline with instructions on alphagov/covid-engineering#1001


junaidsharif:
Not needed anymore as original implementation of #1001 doest need this anymore. Closing PR/branch

PR #180, opened 2020-11-23
Adding the isle of scilly as a test tier locations for gatling tests

whpearson:
As the yaml file defining tiers will be dynamic we cannot rely on it for
performance test purposes.

This change means we will concatenate a yaml file that specifies the
isle of scilly on to the end of the tier lad file is in lockdown. This
was picked as it is less likely to be in lockdown and will therefore not
interfere with legitimate entries but also will not likely be used for
hand testing purposes.

This also changes all the gatling test users to use an isle of scilly
postcode.

Very much a proof of concept.

### DIFF
--- a/loadtests/src/test/resources/data/svp_submission_data.json
+++ b/loadtests/src/test/resources/data/svp_submission_data.json
@@ -2,7 +2,7 @@
   {
     "applying_on_own_behalf": "1",
     "nhs_login": "0",
-    "postcode": "BB1 1TA",
+    "postcode": "TR21 0LW",
     "nhs_letter": "1",
     "nhs_number": "7017309170",
     "first_name": "gatling-test-Matthew",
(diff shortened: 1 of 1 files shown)


whpearson:
To test this do the following. Checkout the branch

Then create a maven docker container as concourse does.
`docker run -v `pwd`:/gatling -ti maven:3.6.3-ibmjava-8-alpine  /bin/bash `

Inside the docker container run the following 

```
cd /gatling/loadtests
mvn gatling:test  -DinjectUsersPerSecond=1         -DinjectDurationSeconds=1         -DpauseBetweenRequestsInSecondsMax=1         -DpauseBetweenRequestsInSecondsMin=1         -DnumberOfRepetitions=1         -Dgatling.simulationClass=svp.SvpSimulationConstantUsersPerSec         -Dgatling.data.writers.0=file         -Dgatling.data.charting.indicators.lowerBound=1000         -Dgatling.data.charting.indicators.higherBound=1500 -Durl=https://gds-shielded-vulnerable-people-service-sandbox.london.cloudapps.digital
```

whpearson:
The real values can be found here https://github.com/alphagov/govuk-shielded-vulnerable-people-service/blob/master/concourse/tasks/gatling-workflow-load-test.yml

PR #183, opened 2020-11-23
Increase CF timeout staging

sanisc:
Staging is timing out when deploying the app so the timeout has been extended to let it complete.


whpearson (approved):
Looks good.

PR #190, opened 2020-11-24
Temp change sandbox pipeline to use feature branch

jk132:
This is for defining sandbox test & deployment tasks
The specific tasks in the branch also will need updating to ensure they build
the app also from the branch code (it's currently set to master)

### DIFF
--- a/concourse/pipeline.yml
+++ b/concourse/pipeline.yml
@@ -22,6 +22,13 @@ resources:
       uri: https://github.com/alphagov/govuk-shielded-vulnerable-people-service
       branch: master
 
+  - name: git-sandbox
+    type: git
+    icon: github-circle
+    source:
+      uri: https://github.com/alphagov/govuk-shielded-vulnerable-people-service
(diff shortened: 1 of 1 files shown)


whpearson:
The trouble is that the e2e job file refer to git-master. So you will need different job files as well or to parameterise them in some way.

E.g. https://github.com/alphagov/govuk-shielded-vulnerable-people-service/blob/master/concourse/tasks/e2e-test.yml#L16

jk132:
> The trouble is that the e2e job file refer to git-master. So you will need different job files as well or to parameterise them in some way.
> 
> E.g. https://github.com/alphagov/govuk-shielded-vulnerable-people-service/blob/master/concourse/tasks/e2e-test.yml#L16

Yeah, plan was to also make the changes in each of the three task files on the feature branch, so sandbox will use those sandbox-specific versions but all other envs will continue using version in master

whpearson:
Ah okay.

whpearson (approved):

PR #191, opened 2020-11-25
Update routing module with postcode tier logic

tal-nagra:
The questions that are displayed to the user are now driven
by the postcode tier (when tiering logic is enabled in app
config).

The answers displayed on the check-your-answers page are also
controlled by the postcode tier (again when enabled).


vulnerable_people_form/integrations/data/local-restrictions-production.yaml
```diff
@@ -329,7 +329,7 @@ E08000022:
 E08000037:
   name: Gateshead Metropolitan Borough Council
   restrictions:
-    - alert_level: 2
+    - alert_level: 4
```
whpearson, line 332: Should we have a tier 4 la in production?

sanisc: this is a mistake - good spot

behave/features/10.e2e_journey_tier_3_manual_address.feature
```diff
@@ -1,5 +1,5 @@
-@e2e_happy_path_no_nhs_login
-Feature: COVID-19 Shielded vulnerable people service - basic e2e user journey - no NHS login
+@feature_postcode_tier
```
jk132, line 1: I suspect I don't fully understand this scenario, journey differences, and split between tags, but why is this not _core_?

jk132 (commented):
Sandeep asked for extra pair of eyes so I've had a look through and all looks generally very sensible (from the perspective of someone who's not really been involved in the web form development to date).
Just a few minor comments/questions. :)

whpearson:
@sanisc It looks like there is no difference between the sandbox and the production environments in terms of postcodes and tiers. We should have a set of known postcodes and tiers that will always be in sandbox and staging, for use in performance and other tests.

vulnerable_people_form/form_pages/shared/session.py
```diff
@@ -191,6 +196,10 @@ def persist_answers_from_session():
     if lives_in_england is not None and lives_in_england == 0:
         lives_in_england = None
 
+    if current_app.is_tiering_logic_enabled and get_postcode_tier() != PostcodeTier.VERY_HIGH_PLUS_SHIELDING.value:
+        # it is not possible to have an answer for basic_care_needs when the tier is v high + shielding
```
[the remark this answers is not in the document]
tal-nagra: Yes @jk132 
There is a typo. We'll get that sorted

Makefile
```diff
@@ -15,12 +15,19 @@ install:
...
+	behave behave/features -k --stop --tags core,e2e_partial_journey_do_you_live_in_england
+endif
 
 smoke_test:
 	@echo "Executing smoke test without submission..."
+ifeq (${TIERING_LOGIC},True)
+	behave behave/features/11.e2e_not_eligible_postcode_tier.feature --stop
+else
 	behave behave/features/5.e2e_not_eligible_postcode.feature --stop
+endif
```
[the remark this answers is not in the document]
sanisc: test_e2e_local is not something that is really used. 

concourse_e2e_with_pipeline_validation_webform_entry - won't need updating as the postcode used is fixed at alert level 4 so the journey for the user is no different from what it is without the feature flag

behave/features/10.e2e_journey_tier_3_manual_address.feature
sanisc: this isn't a core test its part of the feature work, its a test for the journey when the alert its 3 aka very high.

Makefile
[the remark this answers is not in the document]
jk132: Cool thanks

sanisc (approved):

PR #192, opened 2020-11-25
Make separate test step for sandbox

whpearson:
As we want sandbox to be tested before it is deployed and testing
git-master errors and is un-informative, instead test git-sandbox.

This does not rely on the pipeline having been updated using git-master, which
I think is okay


jk132 (dismissed):
Looks sensible I think! Double newline on line 95 but I'm not aware that it's invalid syntax

jk132 (approved):

PR #195, opened 2020-11-26
Update local-restrictions yaml file

sanisc:
The latest local restrictions file for the gov.uk postcode checker has been added.
https://github.com/alphagov/collections/blob/master/lib/local_restrictions/local-restrictions.yaml


niyioyelade (commented):
Looks good to me

samcrang (approved):

PR #196, opened 2020-11-26
Change pipeline to pull tiering flag from concourse

jk132:
Change tiering logic flag in all concourse jobs to pull the value from the
corresponding concourse variable.
This allows the feature to be toggled in each environment by changing the
variable and restaging the app, rather than requiring a code change


whpearson:
Looks good, the mixed casing of the variable names _- jars a bit, but I see why you did it.

whpearson (dismissed):

sanisc:
LGTM

jk132:
As suggested, changed format of params to be more consistent with others (i.e. dash separated between all words).
Changed the params in Concourse to reflect this.

sanisc (approved):

PR #197, opened 2020-11-26
Add gatling test data for the tiering_logic version of the app

jk132:
Add a new test data file which uses a dummy postcode for all entries, 
which the new tiering logic deems to be in very high shielding.
The file is added alongside the old file. It will not currently be used by the
nightly gatling tests, which will continue to use the old version of the app
with the corresponding old test data. It can be used when doing a manual run of
the test, and will become the main file once the tiering_logic version of the
app is released.


whpearson (approved):
Looks good. Having a separate file for now makes sense.

PR #199, opened 2020-11-27
Task/dynamic journey amends

sanisc:
- Update yaml file in non production environments 
- force the basic care needs to be answered when a user changes address 
- add gatling LA into local restrictions yaml file


whpearson (approved):
Changes look good from a code side of things.

PR #200, opened 2020-11-27
Fixed logic for tier 4 prompting basic needs question

sanisc:
When a user changes address from a tier 3 to a tier 4 if they haven't been been asked the basic care needs question they should be when redirected to the view answers page.


whpearson (approved):
Good, but this logic should have a test if possible in another PR.

PR #201, opened 2020-11-27
Default postcode tier to very high plus shielding without tiering logic

whpearson:
As we need to know what questions to ask someone when they return we
need an appropriate postcode tier level to give them, even under
national lockdown.

This sets the value to VERY_HIGH_PLUS_SHIELDING as the questions
currently asked match this level.


samcrang (approved):
I think after this goes in @niyioyelade should be good to update his PR to set previous responses to `VERY_HIGH_PLUS_SHIELDING`.

PR #204, opened 2020-11-30
Add a feature flag to use the tiering logic data

whpearson:
It might make running gatling tests easier if we have to do more before
the 2nd.

Or it might be a model for how to switch it in future.


jk132 (approved):
Looks sensible

PR #212, opened 2020-12-02
Fix returning user tier persistence

sanisc:
Only tier 3 and 4 should be persisted to the database.

### DIFF
--- a/vulnerable_people_form/form_pages/shared/routing.py
+++ b/vulnerable_people_form/form_pages/shared/routing.py
@@ -151,12 +151,12 @@ def update_lives_in_england_referrer(referrer):
         session[SESSION_KEY_LIVES_IN_ENGLAND_REFERRER] = referrer
 
 
-def _get_next_form_url_based_on_postcode_tier(_redirect):
+def _get_next_form_url_based_on_postcode_tier(redirect_url, should_redirect_to_next_form_page=False):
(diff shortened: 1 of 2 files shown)


vulnerable_people_form/form_pages/shared/routing.py
```diff
@@ -151,12 +151,12 @@ def update_lives_in_england_referrer(referrer):
         session[SESSION_KEY_LIVES_IN_ENGLAND_REFERRER] = referrer
 
 
-def _get_next_form_url_based_on_postcode_tier(_redirect):
+def _get_next_form_url_based_on_postcode_tier(redirect_url, should_redirect_to_next_form_page=False):
     postcode_tier = get_postcode_tier()
```
whpearson, line 155: Should this be getting the postcode_tier from the session, or should it make sure that that postcode_tier has been updated? E.g. postcode_tier is populated from tier_at_submission in load_answers_into_session_if_available and so it could theoretically be the old submissions postcode tier being checked here.

sanisc: it will always be the latest one in session for a returning user as the nhs login callback with check the tier at that point and update the session with the correct value

whpearson (approved):
Looks good now I understand what is being changed.

PR #214, opened 2020-12-03
Local development improvements

whpearson:
This should make local development less painful.

It adds the right values in instance/config.py.sample.dev to use the
local ons and disable notify.

It preopulates the database ref_NSPL_postcode_to_lad_code_mapping with
the required values to make e2e pass.

Updates for the README.md

Local e2e adds in the TIERING_LOGIC flag to


instance/config.py.sample.dev
```diff
@@ -35,7 +35,10 @@ AWS_SECRET_ACCESS_KEY = "TEMP"
 
 LOCAL_AWS_ENDPOINT_URL="http://rds-client-api:80"
 
-TIERING_LOGIC=True
+TIERING_LOGIC="True"
+OVERRIDE_ONS_URL="http://ons-test:8000"
```
sanisc, line 39: where is this used in the code?

whpearson: https://github.com/alphagov/govuk-shielded-vulnerable-people-service/blob/8c1c8bba03d34f0f07331f9772ddd010d6205cf8/vulnerable_people_form/integrations/postcode_lookup_helper.py#L87

sanisc: i would make the 2nd part of the url a constant as is duplicated in the helper

vulnerable_people_form/form_pages/shared/answers_enums.py
```diff
@@ -26,7 +26,7 @@ class ApplyingOnOwnBehalfAnswers(EnumWithText):
     NO = 1
 
     __text_values__ = {
-        YES: "Myself",
+        YES: "Myslefssf",
```
sanisc, line 29: revert this change

sanisc (changes requested):

sanisc (approved):

sanisc:
Looks like some good improvements for running ODS locally

PR #216, opened 2020-12-04
Fix missing address bit

niyioyelade:
Add administrative area to town city portion of address.


whpearson:
Do you know if MI uses the town_city field much? Will stats be confused with new data coming in?

niyioyelade:
> Do you know if MI uses the town_city field much? Will stats be confused with new data coming in?

It's worth checking this, I will ask someone from the MI team

niyioyelade:
> Do you know if MI uses the town_city field much? Will stats be confused with new data coming in?

After speaking with MI, I understand most of the MI stuff is postcode driven, so a change to the content of town_city will not make any difference. I also checked this with Andrew Hobson, he is fine with the change as well.

whpearson (approved):
This looks okay, it would be nicer if the faker didn't use the city data twice, so that the test could be sure of the ordering. But that shouldn't hold up the change.

PR #218, opened 2020-12-04
Make changes to send user notifications to backend sqs queue

sh31:
create sqs locally for development


instance/env_to_config_shim.py
```diff
@@ -34,6 +35,7 @@
 
 AWS_RDS_DATABASE_ARN_OVERRIDE = os.environ.get("AWS_RDS_DATABASE_ARN_OVERRIDE")
 AWS_RDS_SECRET_ARN_OVERRIDE = os.environ.get("AWS_RDS_SECRET_ARN_OVERRIDE")
+AWS_SQS_QUEUE_URL = os.environ.get("AWS_SQS_QUEUE_URL")
```
sanisc, line 38: as this is a required field the root __init__.py needs updating with this value

vulnerable_people_form/form_pages/check_your_answers.py
```diff
@@ -7,7 +7,7 @@
 from .shared.routing import route_to_next_form_page, dynamic_back_url
 from .shared.session import persist_answers_from_session, form_answers, get_summary_rows_from_form_answers, \
     get_postcode_tier, is_very_high_plus_shielding_without_basic_care_needs_answer
-from ..integrations import govuk_notify_client, spl_check
+from ..integrations import govuk_notify_client, spl_check, notifications
```
sanisc, line 10: if we are not using the notify client i would remove it from the code base along with the templates

vulnerable_people_form/integrations/notifications.py
```diff
@@ -0,0 +1,66 @@
...
+import json
+
+import boto3
+from botocore.config import Config
+from flask import current_app
+
+from vulnerable_people_form.form_pages.shared.session import form_answers
+from vulnerable_people_form.form_pages.shared.logger_utils import init_logger
+
+logger = logging.getLogger(__name__)
```
sanisc, line 11: can some unit tests be added for this new client

instance/config.py.sample.dev
```diff
@@ -34,11 +34,13 @@ AWS_SECRET_ACCESS_KEY = "TEMP"
...
+LOCAL_SQS_ENDPOINT_URL="http://sqs-client-api:4566"
 
 TIERING_LOGIC="True"
 OVERRIDE_ONS_URL="http://ons-test:8000"
 
 AWS_RDS_DATABASE_ARN_OVERRIDE =  "arn:aws:rds:eu-west-2:123456789012:cluster:dummy"
 AWS_RDS_SECRET_ARN_OVERRIDE = "arn:aws:secretsmanager:eu-west-2:123456789012:secret:dummy"
+AWS_SQS_QUEUE_URL = "http://localhost:4566/000000000000/submissions-message"
 
 NOTIFY_DISABLED="True"
```
sanisc, line 46: if this variable is no longer required then lets get it removed

vulnerable_people_form/form_pages/check_your_answers.py
```diff
@@ -33,8 +33,6 @@ def post_check_your_answers():
     registration_number = persist_answers_from_session()
     session["registration_number"] = registration_number
 
-    is_spl_match = spl_check.check_spl(form_answers()["nhs_number"], form_answers()["date_of_birth"])
-    if "NOTIFY_DISABLED" not in current_app.config:
-        govuk_notify_client.send_notification(registration_number, is_spl_match)
+    notifications.send_message(registration_number)
```
sanisc, line 36: if we no longer do the SPL check can we also remove that from the code base - e.g in the integrations folder

sanisc (changes requested):

sanisc:
the changes/comments requested will also fix the flake8 issues

tests/integrations/test_notification.py
```diff
@@ -0,0 +1,71 @@
...
+        submission_id = "test-registration-number"
+        send_message(submission_id)
+
+        sqs_response = sqs.receive_message(QueueUrl=queue_url)
+        message = json.loads(sqs_response["Messages"][0]["Body"])
+
+        print(f'a: {message}')
+
+        postcode = form_answers()["support_address"]["postcode"]
+        if len(postcode) == 6 and " " not in postcode:
```
whpearson, line 51: Why is this logic needed? If the postcode is set to SW1A 9AJ I can't see why it would ever be 6

sanisc: @sh31 it feels like we are testing the sqs client, we would just want to test the fact that send message behaves as expected

sanisc (commented):
Are the HTML pages/text files still around for the notifications? If so they might as well be removed.

tests/integrations/test_notification.py
thomasridd: I've seen this logic before in `session.py` - it feels like an early version of postcode handling which doesn't actually match up to the actual postcode system but doesn't actually break anything - we may want to pull both

sanisc: @thomasridd the test has been updated since this version

vulnerable_people_form/integrations/notifications.py
```diff
@@ -0,0 +1,66 @@
...
+def get_sqs_client(app=current_app):
+    return boto3.client("sqs", **get_client_kwargs(app))
+
+
+def send_message(registration_number, app=current_app):
+    client = get_sqs_client()
+
+    postcode = form_answers()["support_address"]["postcode"]
+    if postcode[-4] != " ":
+        postcode = f"{postcode[:-3]} {postcode[-3:]}"
```
thomasridd, line 40: https://github.com/alphagov/govuk-shielded-vulnerable-people-service/pull/226 adds a new function `postcode_with_spaces` which would be good to substitute here if it goes in first

vulnerable_people_form/integrations/notifications.py
```diff
@@ -0,0 +1,66 @@
...
+        "address_line_1": form_answers()["support_address"]["building_and_street_line_1"],
+        "address_line_2": form_answers()["support_address"].get("building_and_street_line_2"),
+        "address_town_city": form_answers()["support_address"]["town_city"],
+        "postcode": postcode,
+        "has_someone_to_shop": form_answers().get("do_you_have_someone_to_go_shopping_for_you"),
+        "wants_supermarket_deliveries": form_answers().get("priority_supermarket_deliveries"),
+        "wants_social_care": form_answers().get("basic_care_needs"),
+        "has_set_up_account": form_answers().get("nhs_sub"),
+        "told_to_shield": form_answers()["nhs_letter"]
+    }
```
thomasridd, line 61: Does this mirror part (but not all of) the display logic for the Check Answers page? If so perhaps we should combine the two as a refactor - not a blocker

tests/integrations/test_notification.py
```diff
@@ -0,0 +1,100 @@
...
+        assert message["address_line_2"] == form_answers()["support_address"]["building_and_street_line_2"]
+        assert message["address_town_city"] == form_answers()["support_address"]["town_city"]
+        assert message["postcode"] == form_answers()["support_address"]["postcode"]
+        assert message["has_someone_to_shop"] == form_answers()["do_you_have_someone_to_go_shopping_for_you"]
+        assert message["wants_supermarket_deliveries"] == form_answers()["priority_supermarket_deliveries"]
+        assert message["wants_social_care"] == form_answers()["basic_care_needs"]
+        assert message["has_set_up_account"] == form_answers().get("nhs_sub")
+        assert message["told_to_shield"] == form_answers()["nhs_letter"]
+
+
```
thomasridd, line 69: Could we get a `test_send_message_where_optional_values_are_none` to make explicit expected behaviour when None is going through? (I've seen python handle None in a bunch of odd way)

thomasridd (changes requested):
Also I think you can remove the notifications dependency from `requirements.txt`

I've asked for a bonus test (which will largely be copy paste) but other comments are not blockers

vulnerable_people_form/integrations/notifications.py
sh31: Updated to use this function

tests/integrations/test_notification.py
sh31: Added

vulnerable_people_form/integrations/notifications.py
sh31: Yes, some of the fields that aren't needed for the comms are dropped.

thomasridd (dismissed):
All good from my perspective

sanisc (approved):

PR #222, opened 2020-12-07
Version smoke and e2e tests based on what has been tested for that env

whpearson:
We are about to change the journey quite significantly, putting the
address picker before any other questions.

This means that the behave tests will be out of sync for staging and
production, while the code is being deployed.

In order to make sure the right tests are run against the right
environment, rely on the e2e tests having [passed] for staging and the
code having been deployed for prod tests [passed].

### DIFF
--- a/concourse/pipeline.yml
+++ b/concourse/pipeline.yml
@@ -264,6 +264,7 @@ jobs:
       - get: every-2m
         trigger: true
       - get: git-master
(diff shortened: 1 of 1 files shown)


sanisc:
Seems okay but needs testing once merged.

sanisc (approved):

PR #225, opened 2020-12-08
Update nhs login page content

sanisc:
Provide better content for a returning user who is trying to view their submission

### DIFF
--- a/vulnerable_people_form/form_pages/nhs_login_link.py
+++ b/vulnerable_people_form/form_pages/nhs_login_link.py
@@ -11,4 +11,5 @@ def get_nhs_login_link():
         "nhs-login-link.html",
         previous_path=dynamic_back_url(),
         nhs_login_href=current_app.nhs_oidc_client.get_authorization_url(),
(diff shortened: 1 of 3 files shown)


vulnerable_people_form/form_pages/nhs_login_link.py
```diff
@@ -11,4 +11,5 @@ def get_nhs_login_link():
         "nhs-login-link.html",
         previous_path=dynamic_back_url(),
         nhs_login_href=current_app.nhs_oidc_client.get_authorization_url(),
+        continue_url="/postcode-eligibility"
```
niyioyelade, line 14: I am wondering if continue_url can have a better name to reflect what the url actually is.

niyioyelade (commented):
Generally looks good to me.

vulnerable_people_form/form_pages/nhs_login_link.py
sanisc: the continue_url is for the continue button

niyioyelade (approved):

PR #226, opened 2020-12-08
Content update - Format postcodes

thomasridd:
### What 
Display postcodes with a space

### Why
Postcodes get stored with spaces stripped - this PR inserts space where appropriate for display purposes only

### DIFF
--- a/tests/form_pages/shared/test_form_utils.py
+++ b/tests/form_pages/shared/test_form_utils.py
@@ -178,12 +178,12 @@ def test_postcode_with_spaces_does_ignore_spaces(input_postcode, expected_output
 
 def test_postcode_with_spaces_handles_short_postcodes():
     output = postcode_with_spaces('A')
-    assert output == ''
+    assert output == 'A'
 
     output = postcode_with_spaces('AB')
-    assert output == ''
+    assert output == 'AB'
 
 
 def test_postcode_with_spaces_handles_none():
     output = postcode_with_spaces(None)
-    assert output == ''
+    assert output is None
--- a/tests/form_pages/shared/test_form_utils.py
+++ b/tests/form_pages/shared/test_form_utils.py
@@ -6,7 +6,8 @@ from vulnerable_people_form.form_pages.shared.form_utils import (
     strip_non_digits,
     sanitise_name,
     sanitise_support_address,
-    format_postcode)
+    format_postcode,
+    postcode_with_spaces)
 
 
 @pytest.mark.parametrize("nhs_num", ["", None])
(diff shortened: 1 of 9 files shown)


vulnerable_people_form/form_pages/shared/form_utils.py
```diff
@@ -58,5 +58,13 @@ def format_postcode(postcode):
     return None
 
 
+def postcode_with_spaces(postcode):
+    if postcode is None or len(postcode) < 3:
+        return ''
```
sanisc, line 63: Should this return an empty string? If it is mandatory from where it is called then it would just blow up on submission or further down the line

vulnerable_people_form/form_pages/shared/session.py
```diff
@@ -137,34 +137,14 @@ def get_summary_rows_from_form_answers(excluded_answers=None):
...
-                        "building_and_street_line_1",
-                        "building_and_street_line_2",
-                        "town_city",
-                        "postcode",
-                    ],
-                    answer,
-                )
-            )
+            address_lines = [
+                answer.get('building_and_street_line_1'),
```
sanisc, line 142: i would use the index syntax here for mandatory fields as it should blow up if first line of address, postcode etc is not in the form answers.

`answer['building_and_street_line_1']`

thomasridd: np

vulnerable_people_form/form_pages/shared/form_utils.py
thomasridd: not blowing up is deliberate since this is intended only to be display logic but it would make sense to return something - I'll return `postcode` in the event of the input not being parsable as a postcode

tests/form_pages/shared/test_form_utils.py
```diff
@@ -147,3 +148,42 @@ def test_format_postcode_should_remove_whitespace_characters(postcode, expected_
...
+def test_postcode_with_spaces_handles_short_postcodes():
+    output = postcode_with_spaces('A')
+    assert output == ''
+
+    output = postcode_with_spaces('AB')
+    assert output == ''
+
+
+def test_postcode_with_spaces_handles_none():
+    output = postcode_with_spaces(None)
```
sanisc, line 188: does this test pass now as the changed would return None?

thomasridd: not locally it doesn't - fixed it but we gotta ask why it is passed by CI

thomasridd: ah - so the only CI is flake8 linting - is there a rationale for this and if not shall we fix it?

sanisc (approved):

PR #228, opened 2020-12-09
Update check answers UI

sanisc:
- Update check answers UI to match figma
- Update NHS login clean function to only allow numbers


whpearson (approved):
All the changes make sense and are tested.

PR #229, opened 2020-12-09
994 Catch Glue Failures as part of the gatling test

vijay-p-123:
Change from aws CLI call to a python script that triggers the glue workflow and verify's the jobs have completed
This requires this pull request to be merged 1st https://github.com/alphagov/covid-engineering/pull/1089

### DIFF
--- a/concourse/pipeline.yml
+++ b/concourse/pipeline.yml
@@ -389,23 +389,38 @@ jobs:
         config:
           platform: linux
           image_resource:
-            type: registry-image
+            type: docker-image
             source:
-              repository: governmentpaas/awscli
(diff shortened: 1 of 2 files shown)


jk132 (approved):
Chatted through with Vijay. All looks sensible. Has been tested in separate pipeline up to now and works successfully. Plan now is to merge then run it to ensure it works correctly as part of the pipeline.

sanisc:
> Chatted through with Vijay. All looks sensible. Has been tested in separate pipeline up to now and works successfully. Plan now is to merge then run it to ensure it works correctly as part of the pipeline.

Sounds good

PR #230, opened 2020-12-10
Add pytest to github workflow

thomasridd:


.github/workflows/python-app.yml
```diff
@@ -23,7 +23,19 @@ jobs:
...
       run: |
         python -m pip install --upgrade pip
-        pip install flake8
+        pip install flake8 pytest
+        pip install -r requirements-dev.txt
     - name: Lint with flake8
       run: |
-        flake8 . --count --show-source --statistics
+        # stop the build if there are Python syntax errors or undefined names
+        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
```
sanisc, line 31: there is a .flake8 file in the repo so are these settings needed?

thomasridd: you're right - if it ain't broke

reverted that

.github/workflows/python-app.yml
```diff
@@ -23,7 +23,16 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install flake8
+        pip install flake8 pytest
```
sh31, line 26: Pytest is already included in requirements-dev.txt

sh31: Should flake8 go in there too?

sanisc: might as well add it in makes sense to

thomasridd: cool - I'll remove it

sanisc: can we added flake8 to dequirements-dev.txt makes sense to @thomasridd

thomasridd: why not - makes it more explicit about the version of the linter we are using too

sanisc (approved):

PR #232, opened 2020-12-11
Add leading zeros to DOB on send_message for SPL check.

sh31:
Fix the send_message function to add leading zeros to the day and month of the date_of_birth attribute

This is required by the MySQL stored procedure to check if the person is on the SPL

E.G. "1994/1/1" will throw an error on the spl proc, but "1994/01/01" will succeed

Added test for this scenario

### DIFF
--- a/vulnerable_people_form/integrations/notifications.py
+++ b/vulnerable_people_form/integrations/notifications.py
@@ -5,7 +5,7 @@ import boto3
 from botocore.config import Config
 from flask import current_app
 
-from vulnerable_people_form.form_pages.shared.form_utils import postcode_with_spaces
+from vulnerable_people_form.form_pages.shared.form_utils import postcode_with_spaces, format_date
 from vulnerable_people_form.form_pages.shared.session import form_answers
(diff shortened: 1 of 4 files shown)


vulnerable_people_form/integrations/notifications.py
```diff
@@ -42,7 +42,7 @@ def send_message(registration_number, app=current_app):
     message = {
         "submission_id": registration_number,
         "nhs_number": form_answers()["nhs_number"],
-        "date_of_birth": f"{date_of_birth['year']}/{date_of_birth['month']}/{date_of_birth['day']}",
+        "date_of_birth": f"{date_of_birth['year']}/{date_of_birth['month']:0>2}/{date_of_birth['day']:0>2}",
```
sanisc, line 45: i would create a method out of this and add it to form_utils so you can add a few extra tests around it to make sure it works as expected for different examples

sh31: Created a method. Do the tests look ok?

sanisc (approved):

PR #234, opened 2020-12-14
Update local restrictions yaml with latest changes

sanisc:
Update to latest file from:

https://github.com/alphagov/collections/blob/master/config/local_restrictions.yml


jk132 (approved):
LGTM. I'm assuming the renaming of LA (from  _Shepway District Council_
to _Folkestone and Hythe_) will have no negative impact.

PR #235, opened 2020-12-14
Smoke test submits form

samcrang:
Submit form when running smoke test

We now filter out the magical NHS number that is valid but will never be
assigned to a real person[1]. This means we can submit the form as part
of the smoke tests without worrying that it'll filter down to
LAs/supermarkets.

[1] https://github.com/alphagov/covid-engineering/pull/1086/files#diffd1cb3b92e44f12aca7c91698084340b13ebb56f9e6f7b70c128fd65fc970667fR49

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,8 +18,8 @@ concourse_e2e:
 	behave behave/features -k --stop --tags core,feature_postcode_tier
 
 smoke_test:
-	@echo "Executing smoke test without submission..."
-	behave behave/features/11.e2e_not_eligible_postcode_tier.feature --stop
+	@echo "Executing smoke test with submission..."
+	behave behave/features/12.e2e_journey_tier_3.feature --stop
 
 test_e2e_local:
 	@echo "Executing e2e automated tests against the local environment..."
--- a/Makefile
+++ b/Makefile
@@ -15,27 +15,15 @@ install:
 concourse_e2e:
 	@echo "Executing e2e automated tests against the staging environment..."
 	# execute all scenarios except no submission and backend submissions test ( the '-' character prefix means skip)
-ifeq (${TIERING_LOGIC},True)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -15,27 +15,15 @@ install:
...
 	@echo "Executing e2e automated tests against the staging environment..."
 	# execute all scenarios except no submission and backend submissions test ( the '-' character prefix means skip)
-ifeq (${TIERING_LOGIC},True)
 	behave behave/features -k --stop --tags core,feature_postcode_tier
-else
-	behave behave/features -k --stop --tags core,e2e_partial_journey_do_you_live_in_england
-endif
 
 smoke_test:
 	@echo "Executing smoke test without submission..."
```
junaidsharif, line 21: probably worth amending comment now since it will do a submission.

junaidsharif (changes requested):
LGTM 👍  and pretty much straight forward.
Though probably worth testing locally ?

Makefile
```diff
@@ -15,27 +15,15 @@ install:
 concourse_e2e:
 	@echo "Executing e2e automated tests against the staging environment..."
 	# execute all scenarios except no submission and backend submissions test ( the '-' character prefix means skip)
-ifeq (${TIERING_LOGIC},True)
```
sanisc, line 18: what if we have to revert the feature flag e.g set it to false the the e2e will fail?

samcrang: Will we ever need to go back? I think at this point the tiering stuff has been in production long enough that the feature flagging should be removed. Am I wrong about that?

sanisc: Its a good question i am not sure to be honest my understanding was to keep it so if it did go into a national lockdown it could easily be toggled back but this view of the world may be outdated now

sanisc (dismissed):

sanisc (approved):

PR #236, opened 2020-12-16
Record LA query param using custom dimension in GA

sanisc:
To distinguish between LA and non LA registrations, a query param has been added e.g ?la=1. In order to have visibility over this in Google Analytics, this needs to be sent as a custom dimension

### DIFF
--- a/vulnerable_people_form/static/js/svp-analytics.js
+++ b/vulnerable_people_form/static/js/svp-analytics.js
@@ -2,11 +2,26 @@ window.GOVUK = window.GOVUK || {};
 
 (function (module, cookieSettings, stripPII) {
     "use strict";
+
+    function getQueryParam(param) {
+        var query = window.location.search.substring(1);
+        var params = query.split("&");
+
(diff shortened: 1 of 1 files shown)


thomasridd:
this looks all good to me 
I'm not sure on the Google Analytics side but assuming the intent is to do this `ga('send', 'pageview', { 'dimension3':  laValue});` if there are is an la parameter shout and you've done a sanity check, shout and I'll approve

sanisc:
> this looks all good to me
> I'm not sure on the Google Analytics side but assuming the intent is to do this `ga('send', 'pageview', { 'dimension3': laValue});` if there are is an la parameter and you've done a sanity check, shout and I'll approve

Yes if the la param is present it should send it as a dimension this has been tested in staging and Raluca from the MI sent a screenshot of it working.

Screenshot 2020-12-16 at 12 49 10

thomasridd (approved):
looks good - testing comment was helpful

PR #237, opened 2020-12-16
Validate phone number using formats from Notify codebase

thomasridd:
- add code to validate text contact phone numbers as Notify SMS compatible in the front end
- test against values from Notify test suite

### DIFF
--- a/tests/form_pages/shared/test_validation.py
+++ b/tests/form_pages/shared/test_validation.py
@@ -475,15 +475,9 @@ def test_validate_support_address_should_return_true_when_a_valid_address_is_pro
          "town_city": "town / city too long town / city too long town / city too long town / city too long town / city too long ", # noqa
          "postcode": "LS11BA"
      },
-     [{
-         "form_field": "building_and_street_line_1",
-         "expected_error_msg": "'Address line 1' cannot be longer than 110 characters"
-     },
-         {
-             "form_field": "building_and_street_line_2",
-             "expected_error_msg": "'Address line 2' cannot be longer than 210 characters"
-         },
-         {"form_field": "town_city", "expected_error_msg": "'Town or city' cannot be longer than 50 characters"}])
+     [{"form_field": "building_and_street_line_1", "expected_error_msg": "'Address line 1' cannot be longer than 110 characters"}, # noqa
+      {"form_field": "building_and_street_line_2", "expected_error_msg": "'Address line 2' cannot be longer than 210 characters"}, # noqa
+      {"form_field": "town_city", "expected_error_msg": "'Town or city' cannot be longer than 50 characters"}])
 ])
 def test_validate_support_address_should_return_false_when_an_invalid_address_is_provided(
         test_case_support_address, field_error_messages
--- a/tests/form_pages/shared/test_validation.py
+++ b/tests/form_pages/shared/test_validation.py
@@ -1,8 +1,9 @@
-import pytest
 from unittest.mock import patch
 
-from vulnerable_people_form.form_pages.shared import validation
+import pytest
 from flask import Flask
+
+from vulnerable_people_form.form_pages.shared import validation
 from vulnerable_people_form.form_pages.shared.answers_enums import (
     ApplyingOnOwnBehalfAnswers,
     MedicalConditionsAnswers,
@@ -31,7 +32,7 @@ def test_validate_name_should_return_true_when_first_name_and_surname_entered():
     with patch(
             _FORM_ANSWERS_FUNCTION_FULLY_QUALIFIED_NAME,
             create_form_answers_with_first_name_and_surname), \
-            _current_app.test_request_context() as test_request_ctx:
+         _current_app.test_request_context() as test_request_ctx:
         test_request_ctx.session["form_answers"] = create_form_answers_with_first_name_and_surname()
         is_valid = validation.validate_name()
         assert len(test_request_ctx.session) == 1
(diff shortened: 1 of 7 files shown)


vulnerable_people_form/form_pages/shared/validation.py
```diff
@@ -295,6 +296,28 @@ def validate_phone_number_if_present(section_key, phone_number_key):
...
+        }
+        return False
+
+
+def phone_number_is_valid_for_notify(phone_number):
+    try:
+        validate_uk_phone_number(phone_number)
+        return True
+    except InvalidPhoneError:
+        return False
```
sanisc, line 318: do we want to log at this point saying it was invalid and possibly the reason ?

thomasridd: sure thing

vulnerable_people_form/form_pages/shared/sms_validation.py
```diff
@@ -0,0 +1,48 @@
...
+uk_prefix = '44'
+
+
+class InvalidPhoneError(Exception):
+
+    def __init__(self, message=None):
+        super().__init__(message or 'Not a valid phone number')
+
+
+def validate_uk_phone_number(number):
```
whpearson, line 22: I think this should be called something that indicates it is looking for mobile phone numbers only.

whpearson (changes requested):

vulnerable_people_form/form_pages/shared/validation.py
thomasridd: actually I've just reviewed our logging code and logging validation errors would be something new - fine to do it this one time only if you want but it might be better as part of a larger "extend logging to cover validation" story

vulnerable_people_form/form_pages/shared/sms_validation.py
thomasridd: I have renamed it `validate_notify_compatible_uk_mobile_number`

vulnerable_people_form/form_pages/shared/validation.py
sanisc: i was thinking more like a warning e.g.

`logger.warning(create_log_message(log_event_names["NHS_LOGIN_USER_CONSENT_NOT_GIVEN"],
                                              f"NHS login warning: {error}, error description: {error_description}"))`

thomasridd: Okay - I've done that - the message doesn't include the phone number - didn't know whether it should or shouldn't on account of the PII factor

sanisc: thats fine i think its right not to store anything in splunk of PII

tests/form_pages/shared/test_validation.py
```diff
@@ -459,32 +461,32 @@ def test_validate_support_address_should_return_true_when_a_valid_address_is_pro
...
          "form_field": "building_and_street_line_1",
          "expected_error_msg": "'Address line 1' cannot be longer than 110 characters"
-      },
-      {
-         "form_field": "building_and_street_line_2",
-         "expected_error_msg": "'Address line 2' cannot be longer than 210 characters"
-      },
-      {"form_field": "town_city", "expected_error_msg": "'Town or city' cannot be longer than 50 characters"}])
+     },
+         {
```
sanisc, line 482: this formatting looks off now ?

vulnerable_people_form/form_pages/shared/validation.py
```diff
@@ -295,6 +297,31 @@ def validate_phone_number_if_present(section_key, phone_number_key):
...
+
+
+def phone_number_is_valid_for_notify(phone_number, logger=None):
+    try:
+        validate_notify_compatible_uk_mobile_number(phone_number)
+        return True
+    except InvalidPhoneError as e:
+        if logger:
+            logger.warning(create_log_message(log_event_names["NHS_LOGIN_USER_CONSENT_NOT_GIVEN"],
```
sanisc, line 320: i don't this constant is right - create new one for this scenario

sanisc (changes requested):

tests/form_pages/shared/test_validation.py
thomasridd: fixed that

thomasridd: fixed that up

vulnerable_people_form/form_pages/shared/validation.py
thomasridd: I've made a new one and put it in appropriately

vulnerable_people_form/form_pages/contact_details.py
```diff
@@ -1,13 +1,20 @@
...
 
 from .blueprint import form
+from .shared.logger_utils import init_logger
 from .shared.render import render_template_with_title
 from .shared.routing import route_to_next_form_page, get_back_url_for_contact_details
 from .shared.session import form_answers, get_errors_from_session, request_form
 from .shared.validation import validate_contact_details
 
 
+logger = logging.getLogger(__name__)
```
sanisc, line 14: this logger is never used in this file?

sanisc (changes requested):

vulnerable_people_form/form_pages/contact_details.py
thomasridd: `logger` is used on line 43. `if not validate_contact_details("check_contact_details", logger=logger):`

To fit existing current architecture I have instantiated loggers in the view files. They get passed down to the validation code as parameters

whpearson (dismissed):

sanisc (approved):

PR #238, opened 2020-12-17
Track change answer links and shopping assistance in GA

sanisc:
A requirement from MI to track when a user changes their answers to allow a better understanding of how the user is using the service. In addition, the do-you-have-someone-to-go-shopping-for-you question answers have also been tracked and sent to GA to help understand how people answer this question.


sanisc:
Screenshot 2020-12-17 at 12 55 35

whpearson (approved):
Tested it locally, it is doing the expected behaviour

PR #240, opened 2020-12-17
Update local restrictions file for PROD

sanisc:
Update to latest version of file.

https://github.com/alphagov/collections/blob/master/config/local_restrictions.yml


junaidsharif (approved):
LGTM 👍

PR #241, opened 2020-12-18
fix phone number logic

thomasridd:
- allow phone number for texts to be empty


whpearson (approved):
Looks good, and the logger is working.

whpearson (approved):

PR #247, opened 2020-12-22
Create helper functions for reading the csv file and assessing shielding advice

junaidsharif:
Create helper functions for reading the csv file and assessing shielding advice

This commit assumes that there will be a csv file available with shielding advice.
It creates functions for
- convert uprn/postcode to ladcode
- get shielding advice for ladcode from csv file with columns
 `ladcode,ladname,shielding_advice,start_date,start_time`

It also creates test cases for same.

### DIFF
--- a/tests/integrations/data/local-area-shielding-advice.csv
+++ b/tests/integrations/data/local-area-shielding-advice.csv
@@ -1,0 +1,12 @@
+ladcode,ladname,shielding_advice,start_datetime_utc
+E06000002,Middlesbrough Borough Council,YES,2020-12-02T00:00:01
+E06000002,Middlesbrough Borough Council,NO,2020-12-10T00:00:01
+E06000003,Dummy Council,NO,2020-12-02T00:00:01
+E06000003,Dummy Council,YES,2020-12-10T00:00:01
+E08000012,Liverpool City Council,YES,2020-08-20T00:00:01
+E08000015,Wirral Borough Council,YES,2020-12-02T00:00:01
+E08000015,Wirral Borough Council,NO,2021-12-02T00:00:01
+E06000018,Nottingham City Council,NO,2020-12-02T00:00:01
+E06000018,Nottingham City Council,YES,2020-12-10T00:00:01
+E0800021,Dummy Council,NO,2020-12-02T00:00:01
+E0800021,Dummy Council,YES,2022-12-02T00:00:01
--- a/vulnerable_people_form/integrations/ladcode_shielding_advice_lookup.py
+++ b/vulnerable_people_form/integrations/ladcode_shielding_advice_lookup.py
@@ -1,0 +1,57 @@
+import csv
+from datetime import datetime
+import logging
+
+from vulnerable_people_form.form_pages.shared.logger_utils import create_log_message, log_event_names
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_NATIONAL_SHIELDING_ADVICE = 'NO'
+FIELDNAMES = [
+    'ladcode',
+    'ladname',
+    'shielding_advice',
+    'start_datetime_utc'
+]
+
+
+class LocalAuthorityShielding:
+    la_shielding_advice_data = {}
+
+    def __init__(self, shielding_advice_csv_file: str):
+        self.la_shielding_advice_data = {}
+        valid_applicable_records = []
+
+        with open(shielding_advice_csv_file, 'r') as f:
+            la_shielding_advice = csv.DictReader(f, delimiter=',')
+
+            # Find all non-future records
+            for la in la_shielding_advice:
+                now = datetime.utcnow()
+                if datetime.strptime(la['start_datetime_utc'], '%Y-%m-%dT%H:%M:%S') < now:
+                    valid_applicable_records.append(la)
+            # Sort by date for every ladcode in ascending order
+            sorted_advice = sorted(
+                valid_applicable_records,
+                key=lambda k: (k['ladcode'], k['start_datetime_utc']),
+                reverse=False)
+            # Keep the latest record only.
+            for record in sorted_advice:
+                ladcode = record['ladcode']
+                self.la_shielding_advice_data[ladcode] = record
+
+    def get_shielding_advice_by_ladcode(self, ladcode: str):
+        """
+        fetches the shielding advice status for supplied LAD Code
+        """
+        if ladcode not in self.la_shielding_advice_data:
+            logger.info(create_log_message(
+                log_event_names["LADCODE_NOT_IN_FILE"],
+                f"{ladcode} not in shielding advice file. Using default: {DEFAULT_NATIONAL_SHIELDING_ADVICE}"))
+            return DEFAULT_NATIONAL_SHIELDING_ADVICE
+
+        shielding_advice = self.la_shielding_advice_data[ladcode]["shielding_advice"]
+        logger.info(create_log_message(
+            log_event_names["SHIELDING_ADVICE_FOR_LADCODE_SUCCESS"],
+            f"{ladcode} has Shielding Advice {shielding_advice}"))
+        return shielding_advice
--- a/vulnerable_people_form/integrations/ladcode_shielding_advice_lookup.py
+++ b/vulnerable_people_form/integrations/ladcode_shielding_advice_lookup.py
@@ -20,38 +20,52 @@ class LocalAuthorityShielding:
 
     def __init__(self, shielding_advice_csv_file: str):
         self.la_shielding_advice_data = {}
-        valid_applicable_records = []
 
         with open(shielding_advice_csv_file, 'r') as f:
             la_shielding_advice = csv.DictReader(f, delimiter=',')
 
-            # Find all non-future records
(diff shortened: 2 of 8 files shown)


vulnerable_people_form/integrations/ladcode_shielding_advice_lookup.py
```diff
@@ -0,0 +1,71 @@
...
+FIELDNAMES = [
+    'ladcode',
+    'ladname',
+    'shielding_advice',
+    'start_date',
+    'start_time'
+]
+
+
+def init(shielding_advice_csv_file: str):
```
samcrang, line 23: Can you make this a class so that the state is a bit more obviously encapsulated?

vulnerable_people_form/integrations/ladcode_shielding_advice_lookup.py
```diff
@@ -0,0 +1,71 @@
...
+            valid_applicable_records,
+            key=lambda k: (k['ladcode'], k['start_datetime']),
+            reverse=False)
+        # Keep the latest record only.
+        for record in sorted_advice:
+            ladcode = record['ladcode']
+            la_shielding_advice_data[ladcode] = record
+
+
+def get_shielding_advice_by_ladcode(ladcode: str):
```
samcrang, line 48: Where are the tests for this?

vulnerable_people_form/integrations/ladcode_shielding_advice_lookup.py
```diff
@@ -0,0 +1,71 @@
...
+
+    shielding_advice = la_shielding_advice_data[ladcode]["shielding_advice"]
+    logger.info(create_log_message(
+        log_event_names["SHIELDING_ADVICE_FOR_LADCODE_SUCCESS"],
+        f"{ladcode} has Shielding Advice {shielding_advice}"))
+    return shielding_advice
+
+
+def _get_start_datetime(la_shielding_advice):
+    return pytz.timezone("Europe/London").localize(
```
samcrang, line 66: I think if we control  this file we should decide to store these dates in the file as UTC datetimes and sidestep this weirdness.

tests/integrations/data/local-area-shielding-advice.csv
```diff
@@ -0,0 +1,8 @@
+ladcode,ladname,shielding_advice,start_date,start_time
```
samcrang, line 1: If we're controlling the file can we use a `start_datetime` in ISO8601?

vulnerable_people_form/integrations/ladcode_shielding_advice_lookup.py
```diff
@@ -0,0 +1,71 @@
...
+    'ladcode',
+    'ladname',
+    'shielding_advice',
+    'start_date',
+    'start_time'
+]
+
+
+def init(shielding_advice_csv_file: str):
+    global la_shielding_advice_data
```
samcrang, line 24: 😱

vulnerable_people_form/integrations/ladcode_shielding_advice_lookup.py
junaidsharif: Regarding this, I thin Im inclining towards leaving it as-is. 
With confirmations in place (for new file) and RDS loading which we'll have to do, this whole module will be de-commissioned pretty quickly (as in a week or 2 tops).

Happy to take your lead though if you think we should do this.

samcrang: I really don't like that it means we have to introduce `global`s—I think every bone in my developer body tells me not to do this.

vulnerable_people_form/integrations/ladcode_shielding_advice_lookup.py
junaidsharif: Sure things... we can do that... Though I suspect that the file we receive will be British time.... Just wondering if we should check whether we can get UTC (from source) or not ?

samcrang: I suspect they will be in London time, yes. I think we should check that and store them as UTC though.

tests/integrations/data/local-area-shielding-advice.csv
junaidsharif: 👍

vulnerable_people_form/integrations/ladcode_shielding_advice_lookup.py
junaidsharif: done !

vulnerable_people_form/integrations/ladcode_shielding_advice_lookup.py
junaidsharif: done !

tests/integrations/data/local-area-shielding-advice.csv
junaidsharif: done !

vulnerable_people_form/integrations/ladcode_shielding_advice_lookup.py
junaidsharif: and fixed as well :)

vulnerable_people_form/integrations/ladcode_shielding_advice_lookup.py
```diff
@@ -0,0 +1,57 @@
...
+        self.la_shielding_advice_data = {}
+        valid_applicable_records = []
+
+        with open(shielding_advice_csv_file, 'r') as f:
+            la_shielding_advice = csv.DictReader(f, delimiter=',')
+
+            # Find all non-future records
+            for la in la_shielding_advice:
+                now = datetime.utcnow()
+                if datetime.strptime(la['start_datetime_utc'], '%Y-%m-%dT%H:%M:%S') < now:
```
samcrang, line 31: Is this right? I don't think this will do the right when the shielding status changes at midnight? I don't think we should be called `__init__` on every request.

junaidsharif: as Discussed, this is now changed . Essentialy 
- Init only reads and sort the file
- Filtering of latest valid record happens with every call to `get_shielding_advice_by_ladcode`. So when time changes, it will still work without needing to initialize the class and/or session

samcrang (approved):

tests/integrations/data/local-area-shielding-advice.csv
```diff
@@ -0,0 +1,12 @@
...
+E06000002,Middlesbrough Borough Council,NO,2020-12-10T00:00:01
+E06000003,Dummy Council,NO,2020-12-02T00:00:01
+E06000003,Dummy Council,YES,2020-12-10T00:00:01
+E08000012,Liverpool City Council,YES,2020-08-20T00:00:01
+E08000015,Wirral Borough Council,YES,2020-12-02T00:00:01
+E08000015,Wirral Borough Council,NO,2021-12-02T00:00:01
+E06000018,Nottingham City Council,NO,2020-12-02T00:00:01
+E06000018,Nottingham City Council,YES,2020-12-10T00:00:01
+E0800021,Dummy Council,NO,2020-12-02T00:00:01
+E0800021,Dummy Council,YES,2022-12-02T00:00:01
```
samcrang, line 12: I've been using `freezegun` to test these data in the future type things. Although I guess we might hope that if this service is still around at the end of 2022 that there are bigger things to worry about.

vulnerable_people_form/integrations/location_eligibility.py
```diff
@@ -75,3 +74,53 @@ def is_postcode_in_england(postcode):
                                    f"Postcode: {postcode} "))
 
     return is_postcode_in_england
+
+
+def get_ladcode_from_postcode(postcode):
```
sanisc, line 79: is there any test data required for this to work locally e.g see the following file https://github.com/alphagov/govuk-shielded-vulnerable-people-service/blob/master/scripts/prepopulate.sql - Does this need updating?

junaidsharif: not really, I dont think so... this piece of work assumes data stored in front end... for test, i created data in `tests/integrations/data/local-area-shielding-advice.csv` which we should be able to use locally as well.. unless you are looking for something more explicit...

We will though need test data same as NSPL. ONSUD or other when we move this functionality to back end. Not sure when that will be.

tests/integrations/data/local-area-shielding-advice.csv
junaidsharif: tbh, i am hoping that this whole integration should be moved to backend by mid/end of January... otherwise its just additional work we will need doing keeping data consistent for front end and (inevitably) MI.

vulnerable_people_form/integrations/location_eligibility.py
samcrang: I think our goal for this story was to lay the ground work as best we could to handle shielding levels in different LAs. We don't currently have "real" data so I think for the time being it makes sense not to have any. This will mean this code in this PR is effectively dead code until we get some real data but I think that's fine.

PR #248, opened 2020-12-23
Fix issue with postcode not been checked if in England

sanisc:
Fixed issue with the wrong postcode not eligible screen appearing. Also added automated test for postcode not in england scenario


whpearson (approved):
Looks good

sh31 (approved):

PR #251, opened 2020-12-23
Add postcode overrides for tests

sh31:
Update get_addresses_from_postcode to return an address for a postcode in the

POSTCODE_TIER_OVERRIDE. The street number is set as the tier from the POSTCODE_TIER_OVERRIDE variable.

Update set_is_postcode_in_england to return True for postcodes in POSTCODE_TIER_OVERRIDE
Update set_location_tier to set tier from value in the POSTCODE_TIER_OVERRIDE app config variable

### DIFF
--- a/concourse/pipeline.yml
+++ b/concourse/pipeline.yml
@@ -104,6 +104,7 @@ jobs:
           GA_CROSS_DOMAIN_ID: UA-145652997-7
           ENVIRONMENT: SANDBOX
           TIERING_LOGIC: ((svp-form/feature-tiering-logic-sandbox))
+          POSTCODE_TIER_OVERRIDE: ((svp-form/postcode-tier-override))
 
   - name: smoke-test-sandbox
     serial: true
@@ -163,6 +164,7 @@ jobs:
           GA_CROSS_DOMAIN_ID: UA-145652997-7
           ENVIRONMENT: STAGING
           TIERING_LOGIC: ((svp-form/feature-tiering-logic-stg))
+          POSTCODE_TIER_OVERRIDE: ((svp-form/postcode-tier-override))
         on_failure: *slack_alert_on_failure
 
   - name: smoke-test-staging
@@ -224,6 +226,7 @@ jobs:
           GA_CROSS_DOMAIN_ID: UA-145652997-1
           ENVIRONMENT: PRODUCTION
           TIERING_LOGIC: ((svp-form/feature-tiering-logic-prod))
+          POSTCODE_TIER_OVERRIDE: ((svp-form/postcode-tier-override))
         on_failure: *slack_alert_on_failure
 
   - name: smoke-test-prod
--- a/concourse/pipeline.yml
+++ b/concourse/pipeline.yml
@@ -118,6 +118,7 @@ jobs:
           WEB_APP_BASE_URL: "https://gds-shielded-vulnerable-people-service-sandbox.london.cloudapps.digital"
           MESSAGE: "Checks that the application deployed to sandbox is not serving HTTP error codes"
           TIERING_LOGIC: ((svp-form/feature-tiering-logic-sandbox))
+          POSTCODE_TIER_OVERRIDE: ((svp-form/postcode-tier-override))
 
 
   - name: e2e-test-sandbox
@@ -179,6 +180,7 @@ jobs:
           WEB_APP_BASE_URL: "https://gds-shielded-vulnerable-people-service-staging.london.cloudapps.digital"
           MESSAGE: "Checks that the application deployed to staging is not serving HTTP error codes"
           TIERING_LOGIC: ((svp-form/feature-tiering-logic-stg))
+          POSTCODE_TIER_OVERRIDE: ((svp-form/postcode-tier-override))
         on_failure: *slack_alert_on_failure
 
   - name: e2e-test-staging
@@ -256,6 +258,7 @@ jobs:
           WEB_APP_BASE_URL: "https://gds-shielded-vulnerable-people-service-staging.london.cloudapps.digital"
           MESSAGE: "Checks that the application deployed to staging is not serving HTTP error codes"
           TIERING_LOGIC: ((svp-form/feature-tiering-logic-stg))
+          POSTCODE_TIER_OVERRIDE: ((svp-form/postcode-tier-override))
         on_failure: *slack_alert_on_failure
       - task: cronitor-heartbeat
         timeout: 1m
@@ -276,6 +279,7 @@ jobs:
           WEB_APP_BASE_URL: "https://coronavirus-shielding-support.service.gov.uk"
           MESSAGE: "Checks that the application deployed to production is not serving HTTP error codes"
           TIERING_LOGIC: ((svp-form/feature-tiering-logic-prod))
+          POSTCODE_TIER_OVERRIDE: ((svp-form/postcode-tier-override))
         on_failure:
           task: cronitor-fail
           file: git-master/concourse/tasks/cronitor.yml
--- a/instance/config.py.sample.dev
+++ b/instance/config.py.sample.dev
@@ -7,7 +7,7 @@ GA_TRACKING_ID = "TEMP"
 GA_CROSS_DOMAIN_TRACKING_ID = "TEMP"
 SENTRY_DSN = "https://abc@sentry.io/123/321"
 
-POSTCODE_TIER_OVERRIDE = '{"TE22RR" : 2, "TE33RR": 3, "TE44RR":4}'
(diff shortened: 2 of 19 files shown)


concourse/pipeline.yml
```diff
@@ -104,6 +104,7 @@ jobs:
           GA_CROSS_DOMAIN_ID: UA-145652997-7
           ENVIRONMENT: SANDBOX
           TIERING_LOGIC: ((svp-form/feature-tiering-logic-sandbox))
+          POSTCODE_TIER_OVERRIDE: ((svp-form/postcode-tier-override))
```
sanisc, line 107: this value needs to be passed to smoke tests and end 2 end tests as well

vulnerable_people_form/form_pages/shared/session.py
```diff
@@ -341,15 +341,23 @@ def set_form_answers_from_nhs_user_info(nhs_user_info):
 
 
 def set_location_tier(location_tier):
-    session[SESSION_KEY_LOCATION_TIER] = location_tier
+    postcode = session["postcode"]
```
sanisc, line 344: i would do this outside of the session where it is called so for example.

```
def update_is_postcode_in_england(postcode, app):
    postcode_in_england = is_postcode_in_england(postcode)

    if postcode in app.config["POSTCODE_TIER_OVERRIDE"]:
        postcode_in_england = True

    set_is_postcode_in_england(postcode_in_england)
```

vulnerable_people_form/integrations/postcode_lookup_helper.py
```diff
@@ -85,6 +85,16 @@ def entry_is_a_postal_address(result):
 
 
 def get_addresses_from_postcode(postcode):
+    if postcode in current_app.config["POSTCODE_TIER_OVERRIDE"]:
```
sanisc, line 88: i don't think we should be skipping the call to ODS as that defies what the smoke test and e2e should do. Also i would leave this file and do the override where it is called. See the code below of what  i mean for the address_look.py.

vulnerable_people_form/integrations/postcode_lookup_helper.py
```diff
@@ -85,6 +85,16 @@ def entry_is_a_postal_address(result):
...
+        return [{
+            "text": f"{current_app.config['POSTCODE_TIER_OVERRIDE'][postcode]}, Test Lane, City, {postcode}",
```
sanisc, line 90:
```
_ADDRESS_ERROR_MESSAGES = {
    postcode_lookup_helper.PostcodeNotFound.__name__: "Could not find postcode, please enter your address manually",
    postcode_lookup_helper.NoAddressesFoundAtPostcode.__name__: "No addresses found for {}, please enter your address manually",
    postcode_lookup_helper.ErrorFindingAddress.__name__: "An error has occurred, please enter your address manually"
}


@form.route("/address-lookup", methods=["GET"])
def get_address_lookup():
    postcode = session.get("postcode")
    if not postcode:
        postcode = form_answers()["support_address"]["postcode"]
    try:
        addresses = postcode_lookup_helper.get_addresses_from_postcode(postcode)
    except (postcode_lookup_helper.PostcodeNotFound, postcode_lookup_helper.NoAddressesFoundAtPostcode, postcode_lookup_helper.ErrorFindingAddress) as e:
        if postcode in current_app.config["POSTCODE_TIER_OVERRIDE"]:
            addresses = _create_test_address(postcode)
        else:
            session["error_items"] = {
                **session.setdefault("error_items", {}),
                "support_address": {"postcode": _ADDRESS_ERROR_MESSAGES[e.__class__.__name__].format(postcode)},
            }
            return redirect("/support-address")

    prev_path = append_querystring_params("/postcode-eligibility")

    return render_template_with_title(
        "address-lookup.html",
        previous_path=prev_path,
        postcode=postcode,
        addresses=addresses,
        **get_errors_from_session("postcode"),
    )


def _create_test_address(postcode):
    return [{
        "text": f"{current_app.config['POSTCODE_TIER_OVERRIDE'][postcode]}, Test Lane, City, {postcode}",
        "value": json.dumps({"uprn": 10000000,
                             "town_city": "Test",
                             "postcode": postcode,
                             "building_and_street_line_1": f"{current_app.config['POSTCODE_TIER_OVERRIDE'][postcode]} Test Lane",
                             # noqa
                             "building_and_street_line_2": ""})
    }]
```

concourse/pipeline.yml
sanisc: also the behave tests need updating to read that value and inject it as these are supposed to secret values

instance/config.py.sample.dev
```diff
@@ -7,6 +7,7 @@ GA_TRACKING_ID = "TEMP"
 GA_CROSS_DOMAIN_TRACKING_ID = "TEMP"
 SENTRY_DSN = "https://abc@sentry.io/123/321"
 
+POSTCODE_TIER_OVERRIDE = '{"TE11RR":1, "TE22RR" : 2, "TE33RR": 3, "TE44RR":4}'
```
sanisc, line 10: tier 1 and 2 are treated the same just keep tier 2 in this collection

sanisc (changes requested):

vulnerable_people_form/form_pages/shared/session.py
```diff
@@ -352,7 +352,11 @@ def get_location_tier():
 
 
 def set_is_postcode_in_england(is_postcode_in_england):
-    session[SESSION_KEY_IS_POSTCODE_IN_ENGLAND] = is_postcode_in_england
+    postcode = session["postcode"]
```
sanisc, line 355: i would move this logic in the location_tier.py just like the above change made for `update_is_postcode_in_england`

instance/config.py.sample.dev
```diff
@@ -7,6 +7,7 @@ GA_TRACKING_ID = "TEMP"
 GA_CROSS_DOMAIN_TRACKING_ID = "TEMP"
 SENTRY_DSN = "https://abc@sentry.io/123/321"
 
+POSTCODE_TIER_OVERRIDE = '{"TE22RR" : 2, "TE33RR": 3, "TE44RR":4}'
```
sanisc, line 10: the values in concourse should be different to this to avoid any leakage

behave/features/5.e2e_not_eligible_postcode.feature
```diff
@@ -20,10 +20,4 @@ Feature: COVID-19 Shielded vulnerable people service - partial user journey - in
         Given I am on the "postcode-eligibility" page
         When I give the "#postcode" field the value "QJ5 7VC"
         And I submit the form
-        Then I am redirected to the "do-you-live-in-england" page
```
sanisc, line 23: i would revert this change as this is for when the tiering logic is set to false in the app

sanisc (changes requested):

whpearson (changes requested):

vulnerable_people_form/form_pages/shared/location_tier.py
```diff
@@ -6,19 +6,27 @@
...
+
+    if postcode in app.config["POSTCODE_TIER_OVERRIDE"]:
+        postcode_in_england = True
+
     set_is_postcode_in_england(postcode_in_england)
 
 
 def update_location_tier_by_uprn(uprn, app):
-    if app.is_tiering_logic_enabled:
+    if uprn == 999:
```
whpearson, line 17: You could get the postcode from session here and do the logic on that. Which would enable us to set different tiers from uprns

whpearson:
It'd be nice to add  `- POSTCODE_TIER_OVERRIDE` to https://github.com/alphagov/govuk-shielded-vulnerable-people-service/blob/master/docker-compose.yml#L10 so that the environment variable was picked up when running e2e locally.

whpearson (approved):
This looks good locally, code makes sense. 

Only thing left before merging is getting the concourse values right.

whpearson:
I've just noticed that behave/features/e2e_test_features/01.simple_web_submission_entry.feature should probably change or be removed.

whpearson (approved):
Looks good and tested locally.

vulnerable_people_form/form_pages/address_lookup.py
```diff
@@ -54,6 +57,17 @@ def get_address_lookup():
     )
 
 
+def _create_test_address(postcode):
+    return [{
+        "text": f"{current_app.config['POSTCODE_TIER_OVERRIDE'][postcode]}, Test Lane, City, {postcode}",
+        "value": json.dumps({"uprn": 999,
```
sanisc, line 63: do we care about showing this uprn in plain text?

PR #253, opened 2020-12-24
Remove content about limited Christmas supermarket deliveries

samcrang:


vulnerable_people_form/templates/priority-supermarket-deliveries.html
```diff
@@ -15,10 +15,6 @@
         </div>
         <div class=govuk-warning-text>
```
sanisc, line 16: i think you want to remove the outer div with the warning text class otherwise the UI will just show an icon with no text

sanisc (changes requested):

vulnerable_people_form/templates/priority-supermarket-deliveries.html
samcrang: oh sorry, you're right

sanisc (approved):

PR #256, opened 2020-12-30
Add shielding advice CSV

samcrang:
Add shielding advice CSV

This can be generated from GOV.UK's data[1] using the following one-liner.

```
{ echo "ladcode,ladname,shielding_advice,start_datetime_utc";  yq -r '.  | to_entries | .[] | {"key": .key, "name": .value.name, "restrictions": .value.restrictions[]} | select(.restrictions.alert_level == 4) | [.key, .name, "YES", .restrictions.start_date + "T" + .restrictions.start_time + ":00"] | @csv' local_restrictions.yml | csvformat
```

This assumes:

- people in an LA should shield if and only if they're in tier 4
- local time is currently the same as UTC

[1] https://github.com/alphagov/collections/blob/master/config/local_restrictions.yml

### DIFF
--- a/tests/form_pages/shared/test_location_tier.py
+++ b/tests/form_pages/shared/test_location_tier.py
@@ -1,8 +1,9 @@
-from unittest.mock import patch
+from unittest.mock import patch, Mock
 from flask import Flask
 import pytest
 
 from vulnerable_people_form.form_pages.shared.constants import PostcodeTier, PostcodeTierStatus
+from vulnerable_people_form.form_pages.shared.answers_enums import YesNoAnswers
 from vulnerable_people_form.form_pages.shared.location_tier import (
     update_location_tier_by_postcode,
     update_location_tier_by_uprn,
@@ -27,16 +28,26 @@ def test_update_location_tier_by_uprn_should_not_update_when_tiering_logic_disab
 def test_update_location_tier_by_uprn_should_update_session_when_tiering_logic_enabled():
     try:
         uprn = 110000
+        la = "E06000008"
         location_tier = PostcodeTier.MEDIUM.value
+        shielding_advice = YesNoAnswers.YES.value
         _current_app.is_tiering_logic_enabled = True
+        _current_app.shielding_advice = Mock()
+        _current_app.shielding_advice.is_la_shielding = lambda ladcode: shielding_advice
         with _current_app.app_context(), \
-            patch("vulnerable_people_form.form_pages.shared.location_tier.get_uprn_tier",
-                  return_value=location_tier) as mock_get_location_tier, \
-            patch("vulnerable_people_form.form_pages.shared.location_tier.set_location_tier") \
-                as mock_set_location_tier:
+             patch("vulnerable_people_form.form_pages.shared.location_tier.get_uprn_tier",
+                   return_value=location_tier) as mock_get_location_tier, \
+             patch("vulnerable_people_form.form_pages.shared.location_tier.get_ladcode_from_uprn",
+                   return_value=la) as mock_get_ladcode_from_uprn, \
+             patch("vulnerable_people_form.form_pages.shared.location_tier.set_location_tier") \
+                as mock_set_location_tier,  \
+             patch("vulnerable_people_form.form_pages.shared.location_tier.set_shielding_advice") \
+                as mock_set_shielding_advice:
             update_location_tier_by_uprn(uprn, _current_app)
+            mock_get_ladcode_from_uprn.assert_called_once()
             mock_get_location_tier.assert_called_once_with(uprn)
             mock_set_location_tier.assert_called_once_with(location_tier)
+            mock_set_shielding_advice.assert_called_once_with(shielding_advice)
     finally:
         _current_app.is_tiering_logic_enabled = False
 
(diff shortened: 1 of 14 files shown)


tests/form_pages/shared/test_location_tier.py
```diff
@@ -52,18 +63,29 @@ def test_update_location_tier_by_postcode_should_not_update_when_tiering_logic_d
...
+                as mock_set_shielding_advice:
+
             update_location_tier_by_postcode(postcode, _current_app)
+            mock_get_ladcode_from_postcode.assert_called_once()
             mock_get_location_tier.assert_called_once_with(postcode)
             mock_set_location_tier.assert_called_once_with(location_tier)
+            mock_set_shielding_advice.assert_called_once_with(shielding_advice)
     finally:
         _current_app.is_tiering_logic_enabled = False
+        del _current_app.shielding_advice
```
sanisc, line 88: whats the reasoning for deleting the shielding advice off the conttext?

samcrang: because the previous line does something similar, I'm not sure I really understand why it's needed either.

samcrang: @sanisc I can take it out, I was trying to keep things consistent with what was already there.

sanisc: the previous line is there so that tiering logic is false by default as the current app i think is kept for the lifecycle of all tests. I think the `del _current_app.shielding_advice` can be removed as it is not used in any other tests so its safe to just leave it always on the current app @samcrang

vulnerable_people_form/integrations/data/local-area-shielding-advice-development.csv
```diff
@@ -0,0 +1,2 @@
+ladcode,ladname,shielding_advice,start_datetime_utc
```
sanisc, line 1: i think we will need another file for the sandbox env

vulnerable_people_form/integrations/data/local-area-shielding-advice-staging.csv
```diff
@@ -0,0 +1,2 @@
+ladcode,ladname,shielding_advice,start_datetime_utc
```
sanisc, line 1: it might be worth keeping staging the same as prod as it can help with testing different areas etc

whpearson:
Which card will deal with adding shielding for specific test postcodes (for e2e tests and the like)? I'm guessing that is when we add in a specific journey? 
If that isn't being done in this card it might be worth adding data to the prepopulate.sql so that it doesn't error, when doing local development (although I think the new postcodes have not been fixed for that yet)
https://github.com/alphagov/govuk-shielded-vulnerable-people-service/blob/master/scripts/prepopulate.sql

samcrang:
@whpearson I'm not sure I understand.

I'm hoping this story will let us start storing whether people should be shielding or not (at the moment that's everyone in tier 4). I had hoped that changing the journey would come as part of separate story.

What things need to change to `prepoulate.sql` in order for things not to error? I'm not sure I understand why this changes breaks things.

samcrang:
https://trello.com/c/kPhX1kVZ/1070-only-display-the-support-question-to-the-users-who-are-shielding this is the story that i'm expected to actually change the way the journey works. It talks about a stored procedure but we're going to use the methods implemented in Python for the time being.

whpearson:
@samcrang I'm trying to run `TIERING_LOGIC=True make test_e2e_local` I think it was previous work that switched out the current postcodes such that prepopulate.sql no longer is useful.

whpearson:
I'll prepare a separate PR to fix preopulate.sql.

whpearson:
Trying to run things locally I get
`botocore.exceptions.ParamValidationError: Parameter validation failed:
Invalid type for parameter parameters[0].value.stringValue, value: 10000001, type: <class 'int'>, valid types: <class 'str'>`

This is because `get_ladcode_from_uprn` expects a string and it is being passed an int.

https://github.com/alphagov/govuk-shielded-vulnerable-people-service/blob/master/vulnerable_people_form/integrations/location_eligibility.py#L106

vulnerable_people_form/form_pages/shared/location_tier.py
```diff
@@ -13,6 +22,8 @@ def update_location_tier_by_uprn(uprn, app):
     if app.is_tiering_logic_enabled:
         location_tier = get_uprn_tier(uprn)
         set_location_tier(location_tier)
+        lad_code = get_ladcode_from_uprn(uprn)
+        set_shielding_advice(app.shielding_advice(lad_code))
```
whpearson, line 26: This gives 
```
File "/app/vulnerable_people_form/form_pages/shared/location_tier.py", line 26, in update_location_tier_by_uprn
set_shielding_advice(app.shielding_advice(lad_code))
TypeError: 'LocalAuthorityShielding' object is not callable
```

vulnerable_people_form/form_pages/shared/location_tier.py
```diff
@@ -21,6 +32,8 @@ def update_location_tier_by_postcode(postcode, app):
     if app.is_tiering_logic_enabled:
         location_tier = get_postcode_tier(postcode)
         set_location_tier(location_tier)
+        lad_code = get_ladcode_from_postcode(postcode)
+        set_shielding_advice(app.shielding_advice(lad_code))
```
whpearson, line 36: I suspect it will have the same error as above.

whpearson (changes requested):

whpearson:
https://github.com/alphagov/govuk-shielded-vulnerable-people-service/blob/f06a402c9149fda34cf98dcc0228739c74118e05/vulnerable_people_form/integrations/persistence.py#L274 is expecting an int parameter and getting a string, e.g. YES or NO.

samcrang:
@sanisc @whpearson `make test_e2e_local` is now green and I've included this PR so that it has #264 in it too.

Sorry about the crappyness of this PR in general.

whpearson (approved):

whpearson:
Tested locally, changes look good.

PR #260, opened 2020-12-31
Only load submission if come via nhs number page

sanisc:
Submissions as part of registration should only replayed to the user if they are at the nhs number stage of the journey


vijay-p-123 (commented):
LGTM

vijay-p-123 (approved):

PR #266, opened 2021-01-07
Use 0/1 to represent False/True for storing shielding advice

samcrang:
Without his change we see these errors:

```
An error occurred (SerializationException) when calling the ExecuteStatement operation: TRUE_VALUE is not an expected Json type
```

### DIFF
--- a/vulnerable_people_form/form_pages/shared/location_tier.py
+++ b/vulnerable_people_form/form_pages/shared/location_tier.py
@@ -22,8 +22,12 @@ def update_location_tier_by_uprn(uprn, app):
     if app.is_tiering_logic_enabled:
         location_tier = get_uprn_tier(uprn)
         set_location_tier(location_tier)
+
         lad_code = get_ladcode_from_uprn(uprn)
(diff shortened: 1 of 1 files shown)


vulnerable_people_form/form_pages/shared/location_tier.py
```diff
@@ -22,8 +22,12 @@ def update_location_tier_by_uprn(uprn, app):
     if app.is_tiering_logic_enabled:
         location_tier = get_uprn_tier(uprn)
         set_location_tier(location_tier)
+
         lad_code = get_ladcode_from_uprn(uprn)
```
sanisc, line 26: does it make sense to move this into a method as its the same logic in 2 places?

whpearson (approved):

whpearson:
Makes sense.

vulnerable_people_form/form_pages/shared/location_tier.py
samcrang: Maybe, but I'm not that keen to refactor unless it starts getting used more widely.

PR #270, opened 2021-01-08
Remove check-contact-details page

thomasridd:
Remove the check-contact-details page
- Update router
- Remove page file
- Remove associated functionality
- Update functional and load tests


whpearson:
Code changes look good. Email address

The html is following govuk email input structure. 

We've removed all the email addresses from the contact details in the tests. This might mean that we are not exercising the sending of confirmation emails in staging. Whether that is something we would notice failures in, is another question, but I thought it worth noting. I think ideally we would have attempts to send emails and some method of validating that they are working, but I am not sure what that looks like.

whpearson (approved):
Email has been added back and the load tests also run fine locally.

PR #276, opened 2021-01-13
Track all link clicks for GA

sanisc:
All a tags will now be tracked and sent to GA for MI purposes


whpearson (approved):
Run locally. I see clicks on the back buttons causing "on-page links" events to be sent.

Clicking on change link still sends a request to google analytics as well. 

It all looks good.

PR #277, opened 2021-01-13
Redact UPRNs being logged

jk132:
To comply with PII regulations, replace UPRNs with <redacted> in any log statements

### DIFF
--- a/vulnerable_people_form/integrations/location_eligibility.py
+++ b/vulnerable_people_form/integrations/location_eligibility.py
@@ -42,7 +42,7 @@ def get_uprn_tier(uprn):
     if not records:
         logger.info(create_log_message(
             log_event_names["UPRN_TIER_NOT_FOUND"],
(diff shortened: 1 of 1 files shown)


vulnerable_people_form/integrations/location_eligibility.py
```diff
@@ -42,7 +42,7 @@ def get_uprn_tier(uprn):
     if not records:
         logger.info(create_log_message(
             log_event_names["UPRN_TIER_NOT_FOUND"],
-            f"No tier found in the database for uprn: {uprn}"
+            f"No tier found in the database for uprn: <redacted>"
```
sanisc, line 45: the f strings are not needed in this case as nothing is being replaced - this should also fix the linting issue

sanisc (changes requested):

sanisc (approved):

PR #281, opened 2021-01-18
Update GA link tracking to handle a with nested svg

sanisc:


whpearson (approved):
The ga event with the NHS login link is giving sensible values with this code.

PR #268, opened 2021-01-07
Change routing for shielding question

whpearson:
This change uses the information from the shielding advice in the
session to determine what questions should be asked.

If shielding advice is 1 it should ask the basic-care-needs question.

Going to wait for https://github.com/alphagov/govuk-shielded-vulnerable-people-service/pull/251 to address local testing changes needed for e2e.

PR #282, opened 2021-01-21
Fix test postcode return journey

sh31:
Fix returning user journey to work with test postcodes.

Based off changes in #268


vulnerable_people_form/form_pages/shared/routing.py
```diff
@@ -298,8 +298,12 @@ def get_redirect_for_returning_user_based_on_tier():
         latest_location_tier = location_eligibility.get_uprn_tier(original_uprn)
         latest_shielding_advice = location_eligibility.get_shielding_advice_by_uprn(original_uprn, current_app)
```
whpearson, line 299: Would there need to be an override for uprn as well in the test data?

sh31: Don't think that's needed. The uprns for the test addresses are stored as NULL in the db. `originial_uprn` should be `None` in this function and shouldn't go into that if block.

sh31:
Merged into #268

PR #283, opened 2021-01-22
Task/remove feature toggle

sanisc:
Feature toggle is no longer needed as changes in tier and shielding can be handled independently now


whpearson:
I think we should clean up TIERING_LOGIC from the load tests as well  https://github.com/alphagov/govuk-shielded-vulnerable-people-service/blob/master/loadtests/src/test/scala/svp/ShieldedVulnerablePeople.scala#L10

whpearson:
Looks good. Maybe rename the test `test_route_to_next_form_page_redirects_to_expected_page_with_tiering_logic` to stop mentioning tiering_logic. Same with the name of the gatling data file?

sanisc:
> Looks good. Maybe rename the test `test_route_to_next_form_page_redirects_to_expected_page_with_tiering_logic` to stop mentioning tiering_logic. Same with the name of the gatling data file (and remove one of them, as there were two)?

Done

jk132 (commented):
All looks very sensible to me (once Will's happy). Shall I remove the _tiering-logic_ concourse variables from concourse?

whpearson (approved):
Looks good. Works locally and tested some returning cases.

PR #292, opened 2021-02-04
submission_tracing to output an anonymised event after writing to the db

jk132:
This introduces a `submission_tracing` integration that has a
`persist_answers_log` method. The aim is to write a stand-alone, detailed but
anonymised, log event to standard out after writing to the database.

The following are taken from the submission:
- submission_reference
- nhs_sub

Additionally, the following are grouped as `submission_details` and shuffled:
- phone_number_calls
- phone_number_texts
- email
- address_line_1
- address_postcode

All of the above goes through a _peppering_ step, the `peppering` method does
the following:
1. takes an object
2. converts it to a string
3. lower cases it
4. removes non-alphanumeric characters
5. SHA256 hashes it with a pepper (`SUBMISSION_TRACING_PEPPER` config item)
6. returns the first 32 characters of the 64 long hex digest

The following are also logged:
- time
- event name (`persist_answers`)
- user agent
- client IP
- boolean of whether details matched the SPL


whpearson:
You'll probably want to add `SUBMISSION_TRACING_PEPPER` to `instance/env_to_config_shim.py`

instance/env_to_config_shim.py
```diff
@@ -37,3 +37,5 @@
 SENTRY_DSN = os.environ.get("SENTRY_DSN")
 
 TIERING_LOGIC = os.environ.get("TIERING_LOGIC")
```
sanisc, line 39: this looks like it needs rebasing

vulnerable_people_form/form_pages/check_your_answers.py
```diff
@@ -30,10 +30,9 @@ def get_check_your_answers():
 
 @form.route("/check-your-answers", methods=["POST"])
 def post_check_your_answers():
-    registration_number = persist_answers_from_session()
+    registration_number, is_spl_match = persist_answers_from_session()
```
sanisc, line 33: is_spl_match can be removed as it doesn't exist anymore

instance/env_to_config_shim.py
```diff
@@ -37,3 +37,5 @@
...
+
+SUBMISSION_TRACING_PEPPER = os.environ.get("SUBMISSION_TRACING_PEPPER")
```
sanisc, line 41: if this value is mandatory, then it needs adding to the init.py file

sanisc (changes requested):

vulnerable_people_form/form_pages/shared/session.py
```diff
@@ -18,7 +18,7 @@
 from .querystring_utils import append_querystring_params
 from .security import sanitise_input
 
-from ...integrations import persistence
+from ...integrations import persistence, spl_check, submission_tracing
```
sanisc, line 21: spl_check can be removed

vulnerable_people_form/integrations/submission_tracing.py
```diff
@@ -0,0 +1,72 @@
...
+            objstr = "".join(e for e in objstr if e.isalnum())
+            # generate hash
+            m.update(f"{pepper}:{objstr}".encode())
+            # return first 32 characters of the 64 character hash hex
+            return m.hexdigest()[:32]
+    return None
+
+
+def persist_answers_log(
+    is_spl_match=None, submission_reference=None, submission_details=[], nhs_sub=None
```
sanisc, line 47: spl match needs removing as it is no longer done on the front end

vulnerable_people_form/integrations/submission_tracing.py
```diff
@@ -0,0 +1,72 @@
...
+            "submission_reference": peppering(submission_reference),
+            "submission_details": [
+                peppering(sd) for sd in submission_details if sd is not None
+            ],
+            "nhs_sub": peppering(nhs_sub),
+        }
+    except Exception as e:
+        log_output = {"time": time.time(), "event": "persist_answers", "error": str(e)}
+
+    print(json.dumps(log_output))
```
sanisc, line 72: use logger instead of print

vulnerable_people_form/integrations/submission_tracing.py
```diff
@@ -0,0 +1,79 @@
...
+import logging
+
+from flask import request, current_app
+
+from vulnerable_people_form.form_pages.shared.logger_utils import init_logger, log_event_names, create_log_message
+
+logger = logging.getLogger(__name__)
+init_logger(logger)
+
+def client_ip():
```
sanisc, line 14: if this are private make them _

sanisc (changes requested):

instance/env_to_config_shim.py
jk132: Done

vulnerable_people_form/form_pages/check_your_answers.py
jk132: Done

vulnerable_people_form/form_pages/shared/session.py
jk132: Done

vulnerable_people_form/integrations/submission_tracing.py
jk132: Done

vulnerable_people_form/integrations/submission_tracing.py
jk132: Done

vulnerable_people_form/integrations/submission_tracing.py
jk132: Done

vulnerable_people_form/integrations/submission_tracing.py
```diff
@@ -0,0 +1,80 @@
...
+        result["platform"] = ua_obj.platform
+        result["browser"] = ua_obj.browser
+        result["version"] = ua_obj.version
+        result["string"] = ua_obj.string
+    return result
+
+
+def _peppering(obj):
+    if obj is not None and current_app:
+        pepper = current_app.config.get("SUBMISSION_TRACING_PEPPER")
```
sanisc, line 38: i don't think this check is required and the pepper should be passed into this method. The main method below handles the validation of the SUBMISSION_TRACING_PEPPER and as this is an internal method it should never be called anywhere else.


  ```
def _peppering(obj, pepper):
    m = hashlib.sha256()
    # convert to string and lower case
    objstr = str(obj).lower()
    # get only the alphanumeric characters
    objstr = "".join(e for e in objstr if e.isalnum())
    # generate hash
    m.update(f"{pepper}:{objstr}".encode())
    # return first 32 characters of the 64 character hash hex
    return m.hexdigest()[:32]
    
```

**Review by sanisc — CHANGES_REQUESTED** [review]

File: vulnerable_people_form/integrations/submission_tracing.py [hunk 9]
```diff
@@ -0,0 +1,80 @@
...
+
+def persist_answers_log(
+    submission_reference=None, submission_details=[], nhs_sub=None
+):
+    log_output = {}
+    try:
+        if not current_app or current_app.config.get("SUBMISSION_TRACING_PEPPER") is None:
+            raise Exception("current_app not available or SUBMISSION_TRACING_PEPPER not set")
+
+        if submission_details != []:
```
sanisc, line 60: replace with `if submission_details` as this handles null check

sanisc (changes requested):

sanisc (approved):

PR #297, opened 2021-02-10
Remove pepper in hashing of submission_reference

jk132:
To avoid difficulty of needing to share the pepper across codebases,
hash the submission_reference without a pepper.
This allows security to cross-reference the hashed value with hashes in
logs from AWS infra


sanisc:
Seems fine and checked locally.

sanisc (approved):

PR #303, opened 2021-02-24
Add GA tracking for selected pages

sanisc:


whpearson (approved):
Tried locally, I can see it trying to request a gif with ec=Page%20interaction in the url when I try and change the basic-care-needs question. So it looks good.

PR #304, opened 2021-02-24
Added shielding advice page for returning user

sanisc:
Following the latest round of testing sessions (round 8), we have agreed that we want to add an interstitial page in the returning NHS login user journey. This page will explain to the user that their shielding advice has now changed and will therefore be asked a new question.


whpearson:
Tested locally (changing the shielding advice for test postcodes for tier 3 without shielding from 0 to 1 and reloading). Looks good.

whpearson (approved):

PR #305, opened 2021-02-25
Add error to title when validation fails - DAC

sanisc:


whpearson (approved):
It now has "Select yes if you have someone who can go shopping for you" if you don't have anything selected and puts error on the page title.

PR #307, opened 2021-03-01
Task/dac report fixes

sanisc:


whpearson (approved):
Changes look okay, links open in new tabs as expected. Harder to check the numeric fields but that can be done in staging with mobile devices.

PR #308, opened 2021-03-02
Update labels based on dac

whpearson:


vulnerable_people_form/templates/postcode-eligibility.html
```diff
@@ -4,13 +4,15 @@
 
 
 {% block form_fields %}
+
     <h1 class="govuk-heading-l govuk-!-margin-top-0 govuk-!-margin-bottom-3">
         <label class="govuk-label govuk-label--l" for="postcode">{{ title_text }}</label>
     </h1>
+    <div class="govuk-body">Answer for the person who needs support</div>
```
sanisc, line 11: This doesn't look like it follows the guidance exactly.

When hint text is provided for a single form field, the hint text must be provided in a separate <span> or <div>. This can then be referred to using aria-describedby on the related <input>. This will ensure that the hint is read out when the user placed focus into the form field. Please refer to the GOV.UK Design System

sanisc (changes requested):

vulnerable_people_form/templates/postcode-eligibility.html
whpearson: Okay, I'll update it.

sanisc (approved):

PR #319, opened 2021-03-12
Set session postcode correctly on changing address

whpearson:
Also validate the address is correct before updating session

### DIFF
--- a/vulnerable_people_form/form_pages/address_lookup.py
+++ b/vulnerable_people_form/form_pages/address_lookup.py
@@ -75,16 +75,20 @@ def _create_test_address(postcode):
 
 @form.route("/address-lookup", methods=["POST"])
 def post_address_lookup():
-    session["form_answers"] = {
-        **session.setdefault("form_answers", {}),
-        "support_address": {**json.loads(request_form()["address"])},
-    }
+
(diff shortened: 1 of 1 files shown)


vulnerable_people_form/form_pages/address_lookup.py
```diff
@@ -75,15 +75,19 @@ def _create_test_address(postcode):
...
     session["error_items"] = {}
 
     if not validate_address_lookup():
         return redirect("/address-lookup")
 
+    session["form_answers"] = {
+        **session.setdefault("form_answers", {}),
+        "support_address": {**json.loads(request_form()["address"])},
+    }
+    postcode = {**json.loads(request_form()["address"])}.get("postcode", None)
```
VivianAllen, line 88: seeing as request_form()["address"] gets parsed into a dict on lines 86 and 88 and 91, is it worth having it done once and stored in a variable?

sanisc: Agree with @VivianAllen

whpearson: Sure

VivianAllen (approved):
lgtm ✅

PR #320, opened 2021-03-14
Add e2e test for churn report

jk132:
Modify e2e test so that population of churn report can be tested
by changing basic-care-needs answer with each test submission.
Changing basic-care-needs each run should result in the entry always appearing in the churn report, due to changed preferences.
Took the straightforward approach of alternating the answer with simply:`new_care_answer = '0' if datetime.now().toordinal() % 2 else '1'` (rather than inspecting the previous submission).
If deemed valueable, further work could be done to test the other types of population of the churn report (e.g. change of address), this can be discussed.

### DIFF
--- a/behave/features/steps/general.py
+++ b/behave/features/steps/general.py
@@ -1,6 +1,7 @@
 import os
 import json
 
+from datetime import datetime
 from behave import then, when, given
 from selenium.webdriver.common.by import By
 from selenium.webdriver.support import expected_conditions
@@ -108,3 +109,11 @@ def set_postcode_value_for_shielding(context, tier):
     html_element = context.browser.find_element_by_css_selector("#postcode")
     html_element.clear()
     html_element.send_keys(postcode)
+
+
+@when('I answer the care needs question differently to last submission for nhs_number "{nhs_number}" in hub "{hub}"')
+def click_basic_care_needs_element(context, nhs_number, hub):
+    # change the care needs response each day
+    new_care_answer = '0' if datetime.now().toordinal() % 2 else '1'
+    html_element = context.browser.find_element_by_css_selector(f".govuk-radios__item input[value='{new_care_answer}']")
+    html_element.click()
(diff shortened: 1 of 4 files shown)


behave/features/steps/general.py
```diff
@@ -108,3 +109,11 @@ def set_postcode_value_for_shielding(context, tier):
     html_element = context.browser.find_element_by_css_selector("#postcode")
     html_element.clear()
     html_element.send_keys(postcode)
+
+
+@when('I answer the care needs question differently to last submission for nhs_number "{nhs_number}" in hub "{hub}"')
+def click_basic_care_needs_element(context, nhs_number, hub):
+    # change the care needs response each day
+    new_care_answer = '0' if datetime.now().toordinal() % 2 else '1'
+    html_element = context.browser.find_element_by_css_selector(f".govuk-radios__item input[value='{new_care_answer}']")
```
VivianAllen, line 118: I've never really poked around in the form html so I might be completely wrong, but this css selector looks a bit generic? Wouldn't it find any govuk radio button on the page?

VivianAllen: would it be possible to use find element by name or id instead?

jk132: Good question. Yes you're correct, it'll return the element with the first use of that selector. So yep if the pages were to gain more radio buttons in future, this could cause the wrong element to be filled out. However in that case presumably the e2e tests would need to change anyway in order to correctly fill out the form with its new additional radio button.

This approach is used across the equivalent steps on all current pages, so will need changing on all if we change here. I don't know if the elements already have ids (i.e. would we need to first change the form pages themselves to add them), but I can check with @sanisc in case we think this is straightforward enough and worth doing. 

However my instinct says that given this is the current approach taken and any breaking changes to the form would need changes to e2e tests anyway, it's probably not worth further investigation/effort at this stage.

Full list of element location methods, for interest: https://selenium-python.readthedocs.io/locating-elements.html

jk132: If you're happy I think it probably makes most sense to proceed with this as is, then depending on Sandeep's view stick a separate card in to change them all should we choose to do so?

VivianAllen: Hadn't clocked its done like this elsewhere, sorry. Yeah as you say, if its always been this and always worked, no real reason to change it now.

VivianAllen (approved):
nice!

PR #333, opened 2021-03-17
Fix pyjwt dependabot

whpearson:
This was failing with str has no method decode.


sh31 (approved):
Looks good. [Since 2.0.0, PyJWT.encode returns strings instead of bytes](https://github.com/jpadilla/pyjwt/pull/513)

PR #337, opened 2021-03-23
Point sandbox at nhs smoke test

whpearson:
Previously e2e tests weren't working because they still referenced
master.

This adds a specific task to e2e-test-sandbox and renames everything.


VivianAllen (approved):
seems good to me

PR #338, opened 2021-03-23
remove unused args from churn report e2e care needs q flip-flop

VivianAllen:
remove unused nhs_number and hub args from click_basic_care_needs_element
e2e step function[1].

[1] govuk-shielded-vulnerable-people-service/behave/features/steps/general.py


VivianAllen:
Hey, seems like these args aren't actually used in the step function they map to, so be neater to remove them?

jk132 (approved):
Good spot; I must've had them in there when I was trying the "lookup previous value" approach and forgotten to remove them after changing to the flip-flop :)

PR #342, opened 2021-03-26
April 1st: Stop search engines indexing the service

thomasridd:
DO NOT MERGE UNTIL SERVICE TAKEN OFFLINE

Add noindex, nofollow tags in base templates for all pages


pipwilson:
I think we _do_ want this in before the service goes offline - we should have had this in all along I think (https://www.gov.uk/service-manual/technology/get-a-domain-name#ensure-users-start-their-journey-on-govuk) and we'd want to let the search engines know about it in advance too.

whpearson (approved):
Changes look good. It covers all the meta_tags definitions and the base template is used everywhere barring non-full page templates like analytics_cookie_banner.html

Looks good when spun up locally.

PR #346, opened 2021-03-29
Fix pipeline definition for smoke test sandbox

whpearson:
It should refer to the sandbox branch rather than master
so that alternate smoke tests can be tested.


VivianAllen (approved):
lgtm 👍

PR #347, opened 2021-04-01
Block access to PaaS app whilst service is closed

samcrang:
This will use the IP safelist app that's deployed by the Support
Service[1]. This is a bit confusing. There is probably some work that
could be done to extract and rename this app so that it's clearer that
it's something shared by multiple apps.

[1] https://github.com/alphagov/covid-engineering/blob/273aa5f811ce8e3cf8cc635a0ba61d2e4a23df40/concourse/support-tool/deploy-ip-safelist.yml#L31-L34


philandstuff (approved):
I mean, given we already use the same route service on the `/metrics` path, I don't see that this PR makes things any more confusing

PR #357, opened 2021-04-13
Make nhs keys not break other tests

whpearson:
In order to make the new environment variables compatible with other
tests that do not use them, make them .get accessed rather than
dictionary lookup.


VivianAllen (approved):
👍

PR #359, opened 2021-04-13
Add a waiter on the submit button being clickable

samcrang:
We now have logs in Concourse for failed smoke test runs. There didn't
appear to be anything that would suggest the HTML responses from the
server are weird. What we seem to see is the submit button is clicked
but nothing happens[1].

I'm having a wild guess that for some reason I don't understand the
button isn't clickable yet.

[1] https://cd.gds-reliability.engineering/teams/covid19/pipelines/svp-form/jobs/continuous-smoke-test-prod/builds/111865


whpearson (approved):
Worth I a go. I'll have a look too.

PR #370, opened 2021-04-21
Adding options and timeout to avoid render issues

whpearson:
We are getting intermittent render issues with chrome

https://stackoverflow.com/questions/48450594/selenium-timed-out-receiving-message-from-renderer

Suggests some options. We are also setting the script timeout too.

If this doesn't work. Probably move to a firefox based headless browser

### DIFF
--- a/behave/features/environment.py
+++ b/behave/features/environment.py
@@ -9,14 +9,22 @@ def before_all(context):
     chrome_options = Options()
     chrome_options.binary = "bin/headless-chromium"
     chrome_options.headless = True
+
     chrome_options.add_argument("--no-sandbox")  # Bypass OS security model
     chrome_options.add_argument("--single-process")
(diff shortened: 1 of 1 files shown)


behave/features/environment.py
```diff
@@ -9,14 +9,22 @@ def before_all(context):
     chrome_options = Options()
     chrome_options.binary = "bin/headless-chromium"
```
VivianAllen, line 10: this is the FOSS version of chrome, right? Reckon that matters, vs full-fat chrome?

VivianAllen (approved):
question about chromium vs chrome, but nothing worth holding up merge for. Worth a shot ✅

behave/features/environment.py
whpearson: Hmm, I'll dig while waiting for errors to occur.

PR #419, opened 2021-08-06
Run tests daily

philandstuff:
This reverts c7d3e0d (which moved lots of things to run weekly) and instead moves to run daily.

The driver is the e2e test with pipeline validation, which implicitly assumes that it runs at least daily, to spot changes that happened the previous day and have been through the pipeline.


whpearson (approved):
Looks good

